Issue #468, opened 2026-07-22
Manifest-driven multi-dictionary delivery: generalize DpdUpdateService, rename repo to cst-dictionaries


fsnow:
Generalize the DPD-only download mechanism into a **manifest-driven, multi-dictionary delivery system**, so new derived dictionaries (DPPN first) ship without an app release. Part of #109.

## The move

Today `DpdUpdateService` downloads one hardcoded asset (`dpd-cst-subset.db`) from one repo. It is already ~90% generic — manifest parse (file + sha256 + two version axes), download-with-progress, SHA-256 verify, decompress, atomic install, Windows `.pending` staging. Only the constant names, the repo coordinates, and the "is this db usable" probe are DPD-specific.

Rather than hardcode N descriptors in the app (which needs an app release per new dictionary), the app reads a **published manifest** — the catalog is data, not code.

## Repo rename

`fsnow/dpd-cst-subset` → **`fsnow/cst-dictionaries`** (decided). The repo generalizes from "the DPD subset" into CST's multi-dictionary distribution point: it hosts the DPD asset (`dpd-cst-subset.db`, name unchanged), DPPN, and future derived assets as **separate** release-asset files — each self-describing its own license — plus the manifest. GitHub auto-redirects the old URL. (Repo rename is the owner's action; app-side URL constant updates here.)

**Rule that survives license-isolation:** keep each dictionary a separate asset file with its own license/attribution metadata — never merge different-licensed data (DPD CC-BY-NC-SA + DPPN PD) into one `.db`. The per-entry manifest license field replaces per-repo isolation.

## The manifest

A single JSON catalog published in the repo (its own release asset or a pinned path), listing every available dictionary:

```
{
  "schemaVersion": 1,
  "dictionaries": [
    {
      "id": "dpd", "displayName": "DPD", "definitionLanguage": "en", "kind": "general",
      "assetFile": "dpd-cst-subset.db.zst", "sha256": "...",
      "sourceVersion": "0.4.20260531", "converterVersion": 3,
      "license": "CC-BY-NC-SA-4.0",
      "attribution": { "title": "Digital Pāḷi Dictionary", "author": "Bhikkhu Bodhirāsa", "url": "..." },
      "minAppVersion": "5.0.0"
    },
    {
      "id": "dppn", "displayName": "DPPN", "definitionLanguage": "en", "kind": "proper-names",
      "assetFile": "dppn.db.zst", "sha256": "...",
[...]

Issue #467, opened 2026-07-22
DPPN dictionary source + the canonical derived-asset pipeline (CST.DictPack)


fsnow:
Add the **Dictionary of Pāli Proper Names (DPPN)** — Malalasekera, revised by Ānandajoti Bhikkhu (June 2025) — as a dictionary source, and build the canonical derived-asset pipeline it is the first user of. Part of #109; depends on the registry spine in #466.

## Why DPPN is the forcing example

It exercises everything a general dictionary source needs that the two flat VRI dictionaries don't:
- **HTML definitions** (its `entry` field is pre-rendered HTML — rendered natively by the #466 WebView meaning pane);
- **an entity/proper-name reference** (people, places, texts, kings, monasteries), i.e. `Kind = ProperNames`, not word glosses;
- **homonym numbers** (`Nāgita 1`, `Nāgita 2`) and **cross-references** (`v.l.`, "also called");
- a headword whose **published form must survive** (capitals, homonym marker) — it can't be reconstructed from the IPE key.

## Ingest the SOURCE, not the consumer bundles

DPPN ships upstream as `source/DPPN.json` — a JSON array of `{ "name": headword-maybe-with-HTML, "entry": HTML definition }` — plus GoldenDict/MDict consumer bundles we will **not** parse (lossy, proprietary-binary, reverse-engineered). We derive from the JSON source, for fidelity, stability, and clear provenance; the upstream `dppn.py` exporter is a ready field-mapping reference.

## The canonical derived-asset schema (DPPN first, every later derived dict after)

A small SQLite the app reads through a generic `SqliteDictionarySource`:

```sql
CREATE TABLE meta(key TEXT PRIMARY KEY, value TEXT);
-- source_id ('dppn'), display_name, definition_language ('en'), kind ('proper-names'),
-- schema_version, source_version, converter_version, license, attribution, author, reviser, homepage
CREATE TABLE entry(
    id         INTEGER PRIMARY KEY,
    ipe_key    TEXT NOT NULL,   -- lowercase IPE, homonym marker stripped, NFC before conversion
    headword   TEXT NOT NULL,   -- verbatim published headword, HTML stripped: "Nāgita 1"
    homonym    INTEGER NOT NULL DEFAULT 0,
    body_html  TEXT NOT NULL    -- sanitized at build time to the closed allowlist (#466)
);
CREATE INDEX ix_entry_key ON entry(ipe_key);
```

[...]

Issue #466, opened 2026-07-21
Multi-source dictionary UI: expose DPD (and beyond) in the dictionary panel, not just the two VRI dictionaries


fsnow:
Promoted from the deferred GUI bullet in #109. This is the first expansion of the dictionary UI beyond the two legacy VRI dictionaries, and it closes the drift where the API served more sources than the panel showed.

> **Status (updated 2026-07-23):** the entire backend for this issue is now **built and merged** — the registry spine, all source adapters, the DPD *and* DPPN sources, and manifest-driven delivery. **What remains is purely UI.** The scope below has been rewritten to be prescriptive so it can be implemented rather than re-designed. The architecture rationale is in the "Architecture settled" comment; this body is the current source of truth.

## The drift (the original problem)

The panel showed exactly two sources — VRI **English** (Childers 1875, #378) and **Hindi** (#380) — because `DictionaryViewModel` binds `IDictionaryService.AvailableLanguages` (the flat-file loader, which knows nothing of DPD/DPPN). The API meanwhile served more, one layer up. Two English sources (Childers and DPD are *both* Pāli→English) also prove the selector must be a **source** picker, not a language toggle — a language list shows one "English" and can't tell them apart.

## What's already done (backend — no UI)

The unifying layer the original scope called for now exists and is live in DI:

- **`IDictionarySource`** (`Id`, `DisplayName`, `DefinitionLanguage`, `Kind`, `IsAvailable`, `Attribution`, `LookupAsync`) and **`DictionarySourceRegistry`** (`.Available` = installed sources only, `.ById(id)`) — the single enumeration point both surfaces read, so they can't drift. (#466 spine, merged in #474)
- **Adapters:** `FlatFileDictionarySource` (en/hi), `DpdDictionarySource` (the old `CompositeDictionaryTool.LookupDpd` body; carries `LemmaId` through), `SqliteDictionarySource` (canonical `CST.Lexicon` assets — DPPN and future imports). `CompositeDictionaryTool` has dissolved; the API tool is now `RegistryDictionaryTool`, which always serves every available source and never reads UI prefs.
- **DPPN** as a real source: canonical lexicon format (#470), read-time key derivation (#472), and the `DppnBuilder` converter (in `cst-dictionaries`).
- **Delivery:** `DpdUpdateService` is now manifest-driven and multi-asset (#468) — a released `dppn.db` installs and the registry advertises it automatically.

[...]

PR #471, opened 2026-07-22
#467: DPPN converter — DPPN.json → canonical lexicon (CST.LexiconTools)

fsnow:
Second implementation PR of the multi-source dictionary work (#467), on the just-merged `CST.Lexicon` format. Adds `CST.LexiconTools` — a **build-time console tool, not referenced by the app** — whose first converter turns the Dictionary of Pāli Proper Names source `DPPN.json` into a canonical lexicon asset.

## Real data drove the design

I fetched and analysed the actual 13,642-record `DPPN.json`, which overturned my synthetic assumptions. The `name` field is a malformed formatted **heading**, not a clean headword: the lemma is the first `<b>` run, an optional bare homonym number follows (`<b>Akatuññatāsutta</b> 01.`), citations sit in `<abbr>`, and the markup is unbalanced. So the converter:

- **`ParseHeadword`** takes the first bold run, folds a trailing homonym number/range back in, and excludes citation chrome.
- **`CleanBody`** reduces a definition to the closed allowlist (`p br i b em strong ul ol li hr abbr`), dropping every other tag and **all** attributes.

DPPN is trusted, fixed, public-domain input with a tiny known tag set, so this is build-time *normalization to our tag set*, not an adversarial sanitizer — hence **no third-party HTML dependency**. A vetted sanitizer belongs on the later untrusted in-app import path.

## Validated end-to-end on the real file

13,642 records → **13,594 entries** (48 section-header stubs skipped), **5,053 homonyms**, and — checked across the whole corpus — zero disallowed tags/attributes survive any body. Lookups round-trip: `jetavana` → 3 homonyms, `anāthapiṇḍika` → its 2 homonyms then the deeper prefix run.

## Fable review — two MEDIUM, real-data defects, all fixed

Fable replicated the regex over the full corpus:
- **M1**: `<` + whitespace (`a < b`) was read as a `<b>` tag, corrupting prose. Fixed to match the HTML spec.
- **M2**: the "no unsafe HTML reaches the WebView" guarantee was a property of the *data*, not the *code* (split-tag reassembly, comments, dangling tags could pass). `CleanBody` now **escapes stray `<`/`>` in every text run**, so the only markup in the output is the bare allowed tags it wrote — safe by construction — plus a build-time post-condition assert. Verified across all 13,594 bodies.
- **L1**: documented + tested the (correct) alternative-title homonym behavior (`Uttiyasutta`), which was untested and comment-contradicted.
- **L2**: `Piyasutta 05-06` (the one ranged heading) stored as `Piyasutta 5`, falsely implying a `Piyasutta 6`; now `Piyasutta 5-6`, with `LexiconKey.SplitHomonym` learning a hyphenated range.
- L3 + nits: entity-decode the stub check; csproj comment; drop unneeded `InternalsVisibleTo`; `Program.cs` error handling.

## Testing

13 converter + 20 lexicon tests, using the **real** malformed shapes; the M2 tests pin the safety property against split-tag/comment/partial-tag inputs. Full suite **929/929**.

## Not in this PR
The built `dppn.db` (13,594 entries) exists locally as a byproduct; it's *delivered* by the manifest work (#468). The runtime `SqliteDictionarySource` + registry (#466 spine) exposes it. #466's UI stays with the Kestrel session.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,6 +30,9 @@ src/CST.Lemma/obj/
 src/CST.Lexicon/bin/
 src/CST.Lexicon/obj/
 
+src/CST.LexiconTools/bin/
+src/CST.LexiconTools/obj/
+
 src/CST.Lucene/bin/
 src/CST.Lucene/obj/
 
--- a/src/CST.Avalonia.Tests/CST.Avalonia.Tests.csproj
+++ b/src/CST.Avalonia.Tests/CST.Avalonia.Tests.csproj
@@ -41,6 +41,7 @@
     <ProjectReference Include="..\CST.Lucene\CST.Lucene.csproj" />
     <ProjectReference Include="..\CST.Lemma\CST.Lemma.csproj" />
     <ProjectReference Include="..\CST.Lexicon\CST.Lexicon.csproj" />
+    <ProjectReference Include="..\CST.LexiconTools\CST.LexiconTools.csproj" />
   </ItemGroup>
 
 </Project>
--- a/src/CST.Avalonia.Tests/Lexicon/DppnConverterTests.cs
+++ b/src/CST.Avalonia.Tests/Lexicon/DppnConverterTests.cs
@@ -1,0 +1,137 @@
+using System;
+using System.IO;
+using System.Linq;
+using CST.Lexicon;
+using CST.LexiconTools;
+using Xunit;
+
+namespace CST.Avalonia.Tests.Lexicon
+{
+    /// <summary>
+    /// The DPPN converter (#467). The `name` field is a malformed formatted heading (lemma in the first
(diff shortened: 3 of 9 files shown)


fsnow:
Superseded, not merged: per the decision to host dictionary converters in the **cst-dictionaries** repo (beside DpdLemmaBuilder), the DPPN converter is being relocated there as a self-contained tool that writes the two-column lexicon format directly (no CST.Lexicon dependency), enabled by the read-time-key rework in #472. This converter's SplitHomonym range fix already landed in #472; its logic (headword extraction, body allowlist filter, all the Fable-review fixes) carries to the cst-dictionaries version. Branch issue-467-dppn-converter left for reference.

Issue #542, opened 2026-07-31
Page anchors resolve one page early at a sutta boundary: the marker that applies to a sutta sits AFTER the first word of its first paragraph


fsnow:
Found in View Source PDF testing: at a **sutta boundary** the app can open the source PDF at the wrong page. The underlying page-anchor algorithm has never been written down anywhere — capturing it here, since it governs View Source, the status-bar page readout, and Go To alike.

## Where the page markers actually are (domain knowledge)

`<pb>` positions in the corpus derive from **footnotes that were in the CST texts** (both the Myanmar and Devanagari editions). They are not synthetic page boundaries.

The consequence at a section heading: **the page marker that applies to a sutta sits after the first word of the first body-text paragraph following the heading.** Verified across the first six suttas of `s0101m.mul.xml` — the shape is identical every time:

```xml
<div id="dn1_1" type="sutta">
<head rend="chapter">१. ब्रह्मजालसुत्तं</head>            <- heading
<p rend="subhead">परिब्बाजककथा</p>                       <- optional subhead
<p rend="bodytext" n="1"><hi rend="paranum">१</hi><hi rend="dot">.</hi> एवं
    <pb ed="M" n="1.0001"/><pb ed="V" n="1.0001"/><pb ed="P" n="1.0001"/><pb ed="T" n="1.0001"/>
    मे सुतं – …
```

Words between the heading and the first following `<pb>`, for the first six DN suttas:

| Sutta | first pb | intervening tokens |
|---|---|---|
| १. ब्रह्मजालसुत्तं | 1.0001 | subhead, paranum, dot, **एवं** |
| २. सामञ्ञफलसुत्तं | 1.0044 | subhead, paranum, dot, **एवं** |
| ३. अम्बट्ठसुत्तं | 1.0082 | paranum, dot, **एवं** |
| ४. सोणदण्डसुत्तं | 1.0104 | subhead, paranum, dot, **एवं** |
| ५. कूटदन्तसुत्तं | 1.0120 | subhead, paranum, dot, **एवं** |
| ६. महालिसुत्तं | 1.0143 | subhead, paranum, dot, **एवं** |

Always the first word (**एवं**), then the markers.

## Why the current algorithm gets it wrong

`getPageReferences` → `findBestAnchor` resolves by **last anchor at or before the current position**:

```js
// Case 1: Linear search for anchor at or before current position
for (var i = 0; i < sortedAnchors.length; i++) {
[...]

PR #544, opened 2026-07-31
#542 follow-up: only a BOOK boundary starts a new page, not every section

fsnow:
Follow-up to #543, which over-fired. Found immediately in testing.

## The regression

#543 looked forward at **every** div boundary. But only a **book** starts a new printed page — a vagga / sutta / samyutta / pannasaka boundary falls mid-page, so the heading there is still on the previous marker's page.

Seen at **`(7) 2. Sukhavaggavaṇṇanā`** (`div an2_2_2`, `type="vagga"`) in `s0402a.att.xml`:

```xml
<p rend="bodytext"> ६४. … <pb ed="M" n="2.0053"/><pb ed="V" n="2.0051"/>
</div>
<div id="an2_2_2" type="vagga">
  <head rend="chapter"> (७) २. सुखवग्गवण्णना        <- still on printed page 53
```

The heading is on **printed Myanmar page 53**; the status bar showed **2.54**.

## Why "book" is the right test

Only book divs begin a new printed page — which is exactly why the salutation appears at them and nowhere else. Across DN/MN/SN/AN:

| div type | count | opens with a salutation |
|---|---:|---:|
| **book** | 66 | **27** |
| vagga | 356 | 0 |
| samyutta | 170 | 0 |
| sutta | 104 | 0 |
| pannasaka | 48 | 0 |
| peyyala | 18 | 0 |

## The discriminator is already in the DOM

Anchors carry the div `id`, and across **all 217 corpus files**:

> `type="book"` ⟺ the id has **no underscore** — **zero violations**.

Every other type (vagga, sutta, samyutta, pannasaka, nipata, peyyala, chapter, intro, vimana) has one. So no new markup, no XSL change, no reseeding.

## Second bug this exposed

Tracking the *last* anchor was insufficient: a book div is immediately followed by its nested pannasaka/vagga divs, so at the title or vagga heading the last anchor is nested and the book start is missed. The Tikanipāta boundary read correctly at the salutation but **not a few lines further down**. Now tracks the most recent **book-level** anchor.

## Safety

**139 of 217 files have no div ids at all** — the rule stays inert and behaviour is unchanged for every unmarked book.

## Verification

Simulation over both boundary types, all passing:

- vagga heading keeps its page (the regression)
- book boundary takes the new page — at the salutation **and** at the nested headings
- closings before the book div keep the old page
- mid-paragraph unchanged · sutta boundary mid-page · top of book · no div markup · end of document

**1003/1003** tests.

## Note
An independent plausibility check was suggested during review: a marker several paragraph anchors away cannot be governing the current position, which rules out 2.54 here on its own. It agrees with this fix. Not implemented, because it is insufficient alone — at this same boundary the next **V** marker *is* in the very next paragraph, so a distance test would still have looked forward and been wrong for VRI pages, while the div test is right for both editions.

### DIFF
--- a/src/CST.Avalonia/Views/BookDisplayView.axaml.cs
+++ b/src/CST.Avalonia/Views/BookDisplayView.axaml.cs
@@ -1228,14 +1228,29 @@ public partial class BookDisplayView : UserControl
                             //
                             // Books with no div markup have no chapter anchors, so sectionStart stays -1 and the
                             // behaviour is exactly as before — no regression where the markup is absent.
-                            var sectionStart = -1;
+                            // Only a BOOK-level div starts a new printed page. A vagga / sutta / samyutta /
+                            // pannasaka boundary falls mid-page, so the heading there is still on the PREVIOUS
+                            // marker's page and 'last at or before' is already correct — looking forward at those
+                            // boundaries jumps a page early. Verified: '(7) 2. Sukhavaggavannana' (div an2_2_2,
+                            // type=vagga) sits on printed Myanmar page 53, and <pb M 2.0053/> precedes it.
+                            //
+                            // Book-level divs are identifiable from the anchor name alone: across the s-corpus
+                            // every type='book' id has NO underscore (88/88) and every other type has one
+                            // (vagga, sutta, samyutta, pannasaka, nipata, peyyala, chapter, intro, vimana).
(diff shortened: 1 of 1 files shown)


fsnow:
Superseded by the precomputed per-edition rule — this version was wrong for VRI at the Sukhavagga boundary (it returned 2.51 where the print reads 2.52). Replaced rather than fixed.

Issue #570, opened 2026-08-07
Feature request: local search (Ctrl+F) in the current book


Antonio-Costanzo:
I would like to request the restoration of a feature that was present in CST 4.1: the ability to use Ctrl+F **to search only within the currently open book**.

In the current version (5.0.0-beta.5), Ctrl+F opens the global search across the entire corpus. While this is useful for broad searches, it makes it much harder to perform simple, quick lookups within the text I'm actively reading.

Here are two specific cases where this feature is essential for me:

1. **Finding a paragraph number**: When I resume reading a book, I often need to quickly jump to a specific paragraph (e.g., "section 15"). The global search makes this unnecessarily complex.
2. **Restoring my position after zooming**: When I increase the text size, the pagination shifts and I lose my exact reading position. My workaround is to copy a line of text before zooming and then use Ctrl+F to paste and find it after zooming, which instantly restores my place. The global search cannot do this for the current view.

Is it possible to bring back the local (in-book) search for Ctrl+F? A local search is a fundamental feature for a reader application.

---

*Maintainer note — cross-reference:* the second motivating case above (losing your reading position when the text size changes) is a **bug**, not something that should need a workaround. Tracked separately as **#571** — on Windows, Ctrl+scroll applies Chromium's layout zoom, which rewraps the text while the app fails to restore the reading position afterwards. Fixing that removes the need for the copy-a-line-then-find ritual, independently of this request.

This issue remains open for the local in-book find itself, which stands on its own. Related: #571 (the position loss), #42 (user-configurable book fonts).

fsnow:
## Feasibility research: CEF has native find-in-page, and it is reachable

Short version: **Chromium's own find is fully implemented in the CEF build we ship** (CEF 120 via CefGlue 120.6099.211 / WebViewControl-Avalonia 3.120.9), including match counts. It needs two small reflection hops to reach, and it has one limitation that needs a product decision.

### The API

| | signature | access |
|---|---|---|
| Search | `CefBrowserHost.Find(string searchText, bool forward, bool matchCase, bool findNext)` | public |
| Clear | `CefBrowserHost.StopFinding(bool clearSelection)` | public |
| Results | `FindHandler.OnFindResult(CefBrowser browser, int identifier, int count, CefRectangle selectionRect, int activeMatchOrdinal, bool finalUpdate)` | `protected abstract` — override |
| Wiring | `BaseCefBrowser.FindHandler { get; set; }` | **public setter on a public type** |

[...]

Issue #434, opened 2026-07-21
Canonical reading-position token (bracketing anchors + fraction) for all capture/persist/restore paths


fsnow:
Adopt one **canonical reading-position token** — the nearest anchors bracketing the viewport top plus the fraction between them — as the single representation used **everywhere** a position is captured, persisted, or restored. Today each path uses a different, coarser mechanism, and each drifts; this unifies them on the most accurate model.

## The model (chosen)

Represent the current position as `{above: A, below: B, fraction}`:
- **Capture:** find the nearest anchor **above** the viewport top (`A`) and the nearest **below** (`B`), across ALL anchor types (page / paragraph / chapter), and record `fraction = (scrollTop − A.position) / (B.position − A.position)`.
- **Restore:** resolve `A` and `B` to their **current** positions and scroll to `A.pos + fraction * (B.pos − A.pos)`.

Interpolating between two bracketing anchors preserves the *relative* reading position even when absolute pixel offsets change — so it is robust to **reflow and different window sizes**, which is exactly what the other schemes get wrong.

> Rejected alternative — **nearest single anchor** (any type): finer than today's page anchor, but restore still lands at that anchor's top edge, so still off by up to the gap to the nearest anchor. Not exact.

## One representation, many consumers

Replace these divergent, drift-prone mechanisms with the single token:

| Consumer | Today | Problem |
|---|---|---|
| **Cross-run serialization** | `ApplicationState.CurrentAnchor` — a single anchor string (`ApplicationState.cs:204`) | coarse; restores to an anchor's top edge on next launch |
| **Script change** | `GetCurrentPageAnchor` → `ScrollToPageAnchor` (page granularity) | worst drift — lands at the page-marker boundary (see original report below) |
| **Tab reattach / recycle** | saved paragraph anchor (`GetCurrentParagraphAnchorAsync`) | paragraph-top edge; reflow shift |
| **Window resize** | native `scrollTop` preserved by the browser | reflow moves the content under that pixel → reading position drifts |
| **Float / unfloat** | re-parent recreates the browser, then restores the saved anchor | same top-edge/reflow drift |

A schema change: `ApplicationState.CurrentAnchor` (string) → the richer token (with a version-gated migration, per the settings/state migration framework).

## Original symptom (what surfaced this)

[...]

Issue #42, opened 2026-06-21
Book-content font face management (user-configurable font faces, injected at render time)


fsnow:
Book content font faces are currently hardcoded in the XSL stylesheets, requiring manual XSL edits to customize (as in CST4, where users really did edit the stylesheets — and those edits do not carry forward, since the XSL is now internal to the app). Make the **font face** user-configurable per Pāli script, separate from the UI per-script font system and from UI-localization fonts (this is the third distinct font system).

## Scope

- **Per-script font face** for book content.
- **Inject the face at render time** rather than generating per-script stylesheets. The 14 stylesheets are now byte-identical except their one `font-family` line, so the intended end state is a **single stylesheet** with the face supplied as a parameter. (Holding that collapse until this injection mechanism exists to consume it.)
- **Live preview** in the book display.
- **Surface the saved zoom level** (#572) alongside the face for each script — a readout plus a reset, not a second editing surface. Editing stays on the keyboard against live text, since judging whether a face looks right at a size can only be done by eye. But Settings is where users look for "book font size", nothing else reveals that zoom is stored **per script** rather than globally, and a panel offering only a face dropdown reads as unfinished. Showing e.g. *Devanagari 160% / Latin 100%* side by side also exposes when two scripts are badly out of step.
- **Report a configured face that isn't installed** (#573). This panel is its natural home: face, zoom, and "this font is gone" all describe the same script.

## Explicitly out of scope: font size

**Size is not configurable here.** It is covered entirely by zoom (#572), which is now the single stored size number, persisted per script:

```
effective book text size = shippedXslSize × zoom[script]
```

A user-editable base size and zoom were two UIs for one number. Zoom wins because the judgment involved — "does this face look right at this size?" — is visual, and can only be made against live text. See the decision record on #572.

The shipped ladder is now uniform across all 14 stylesheets (body 12pt, chapter 18pt, book 21pt, nikaya 24pt).

## Implementation requirement: preserve the reading position

Changing the font face **reflows the text**, moving every anchor. Unless the reading position is carried across, the reader is thrown to a different part of the book on every adjustment — including mid-drag if there is a live preview.

The machinery exists and should be reused. The resize path already snapshots the #434 `ReadingPositionToken` before the reflow and restores it once layout settles:

```
BookDisplayView.axaml.cs  ~609  "Resize started - snapshotted reading position (above=…, below=…, frac=…)"
[...]

PR #591, opened 2026-08-10
Book text zoom: Cmd/Ctrl +/-/0, persisted per script (#572)

### LINKED ISSUE #571: Windows: Ctrl+scroll resizes book text but the reading position is not restored
## Symptom
On Windows, **Ctrl + scroll** (trackpad or wheel) inside a book changes the text size. The text rewraps and the reader loses their place — the viewport stays at the same scroll offset while the content reflows underneath it.

## Why it is Windows-only
Chromium has two distinct zoom models, and the platforms expose different ones:

| | gesture | mechanism | reflow? |
|---|---|---|---|
| **Windows** | Ctrl + scroll | **layout zoom** — changes the CSS pixel ratio, relayouts, rewraps | **yes** |
| **macOS** | pinch | **page-scale zoom** — magnifies the rendered layout like a lens | no |

macOS pinch is harmless: document coordinates and `pageYOffset` are untouched, so nothing in our tracking is affected (verified by hand — scroll position survives a pinch). Only the Windows path reflows, and only it loses the reading position.

## What already works
The injected page script registers an in-page `resize` listener that rebuilds `cstAnchorCache` after a 100 ms settle delay (`BookDisplayView.axaml.cs` ~1559). A layout zoom fires that event, so the **anchor cache is refreshed** and the derived readouts — page-number status bar, Go To — stay correct.

## What is missing
That handler rebuilds the cache but **does not restore the reading position**. Compare the two reflow paths:

| path | trigger | cache rebuild | position preserved |
|---|---|---|---|
| Avalonia control resize (window resized) | `SizeChanged` → resize-settle timer | yes | **yes** — snapshots the #434 token, restores after settle (~609/636) |
| In-page resize (Windows layout zoom) | JS `resize` → `cstAnchorCache.build()` | yes | **no** |

The control-resize path never fires for a zoom: the Avalonia control's size doesn't change, only the page's layout underneath it. And the JS handler never tells C# anything, so the position machinery is never invoked.

## Suggested fix
**Do not try to snapshot on `resize`** — by the time it fires the reflow has already happened, so the pre-reflow position is gone. Use the token that is *already* being kept fresh: the #434 rolling capture refreshes `_lastPositionToken` on every status tick (~200 ms), so a good position is always on hand and at most one tick stale.

So: have the in-page `resize` handler notify C# (the existing `document.title` channel), and once the cache rebuild has settled, call `ScrollToPositionToken(_lastPositionToken)` — exactly what the #31 tab-reattach path already does. Ordering matters: rebuild the cache first, then restore, or the restore resolves against stale positions.

## Notes
- **User evidence:** a beta tester filed #570 asking for an in-book Ctrl+F, and gave as one of two motivating cases "copy a line of text before zooming, then find it after zooming to restore my place". That feature request is partly a workaround for *this* bug.
- **Cross-platform inconsistency:** Windows users can resize book text today via an undocumented gesture; macOS users cannot resize it at all. Managed sizing is #42 — which means #42 is not purely additive on Windows, since it has to reconcile with a zoom that already exists.
- The zoom level's persistence across reloads/sessions is untested and worth checking while fixing this.

Related: #42 (user-configurable book fonts), #434 (reading-position token), #570 (the request that surfaced this), #321 (anchor-cache rebuild after reflow).

### LINKED ISSUE #572: Book text zoom: support ⌘/Ctrl + plus/minus/0 on both platforms
There is currently **no deliberate way to zoom book text**, and what exists is accidental and platform-split:

| platform | gesture | what happens |
|---|---|---|
| macOS | ⌘ + plus/minus | **nothing** |
| macOS | pinch | Chromium *page-scale* zoom — magnifies the rendered layout like a lens, no rewrap (so you must pan horizontally to read) |
| Windows | Ctrl + scroll | Chromium *layout* zoom — genuinely resizes and rewraps the text, but the app does not know it happened (#571) |

So a Windows user can resize book text via an undocumented gesture that loses their reading position, and a macOS user cannot resize it at all. Neither behaviour was chosen.

## What to build
The standard reader/browser bindings, on **both** platforms:

- **⌘/Ctrl + `+`** — zoom in
- **⌘/Ctrl + `-`** — zoom out
- **⌘/Ctrl + `0`** — reset to 100%

This is the **Zoom (⌘+/−/0)** entry on #111's catalogue of behaviours *inherited from CST4's WinForms WebBrowser (embedded IE) control and never explicitly coded* — the same list that produced #112 (printing, shipped) and #570 (find on page). CST4 got zoom free from IE; with CEF it has to be wired deliberately.

## Implementation notes
- **Two possible mechanisms.** (a) Chromium's own zoom via `CefBrowserHost.SetZoomLevel`/`GetZoomLevel` — needs two reflection hops — `WebView.UnderlyingBrowser` (internal getter) then `BaseCefBrowser.UnderlyingBrowser` (protected getter) — after which `GetHost()` is public; zoom levels are logarithmic (`level = log(factor)/log(1.2)`). The same chain is mapped in detail on #570. (b) CSS injection — scale the book's font sizes in the page, which avoids reflection and would share a mechanism with #42's CSS-injection approach. **(a) is the one to use** — see the decision record in the comments. It scales all the absolute `pt` classes for free, which is why today's accidental Ctrl+scroll zoom already renders every heading correctly; (b) would mean overriding ~15 classes or refactoring the stylesheets to relative units first. #42 is now face-only and never touches sizes, so the two mechanisms decouple.
- **Reading position must be preserved across the reflow.** Zooming rewraps the text, so this is the same requirement as #571 and #42 — restore the rolling #434 `ReadingPositionToken` after the anchor-cache rebuild settles. Implementing zoom properly subsumes #571's fix.
- **Bind through the existing plumbing**: `PlatformGesture` (#509) for the per-platform command modifier — ⌘ on macOS, Ctrl on Windows — and the shortcut must survive focus being inside the WebView, which is what `WebViewShortcutRelay` (#562) exists for. Registration points: `SimpleTabbedWindow.RegisterMenuShortcutKeyBindings` and `App.RegisterFloatingWindowShortcuts`.
- **Keyboard gotcha:** the `+` key is `=` unshifted on most layouts. Bind `OemPlus` and `Add` (numpad), and accept the keystroke with or without Shift, or ⌘+ will appear not to work for many users.
- **Decide the scope of a zoom level**: per book tab, or global for all books? And does it persist across relaunch? (Chromium persists zoom per-origin by default, which for our local content may produce surprising carry-over — worth checking rather than inheriting.)
- **Windows Ctrl+scroll** should end up going through the same path rather than remaining a separate uninstrumented one.
- A **View menu** entry alongside the shortcuts would make the feature discoverable — a beta tester has already demonstrated that undiscoverable features get worked around rather than found.

## Relationship to the neighbouring issues
- **#42** — user-configurable book font *faces*. Zoom owns size entirely; there is no separate base-size preference. See the decision record in the comments here.
- **#573** — a configured face can silently vanish, which leaves the zoom stored for that script calibrated for nothing.
- **#571** — the Windows bug where today's accidental zoom loses the reading position. Fixed as part of doing this properly.
- **#570** — the other unwired inherited-IE behaviour (find on page), from the same #111 list.
- **#111** — the catalogue of inherited-WebBrowser behaviours.

fsnow:
Adds a deliberate way to resize book text on both platforms. Until now macOS had none at all, and Windows had only Chromium's undocumented Ctrl+scroll — which reflowed the text and lost the reader's place, because nothing on the C# side knew it had happened (#571). A beta tester built a copy-a-line-then-search ritual around exactly that (#570).

```
effective book text size = shippedXslSize × zoom[script]
```

## The model

Zoom is the **only** stored size control for book content. #574 flattened all 14 stylesheets to one ladder and #42 was narrowed to font *faces*, so there is deliberately no user-editable base size — the two were redundant, and a live control beats typing a number into a dialog for a judgement that can only be made by eye.

Stored **per script**: with the ladder now shared, the only thing zoom compensates for is the metrics of whatever face that script resolves to — the same axis the face setting uses. Switching a book's script switches face and zoom together.

**Book text only.** App chrome keeps following its own font settings.

## Mechanism

`CefBrowserHost.SetZoomLevel`, not CSS injection. It scales every absolute `pt` class proportionally for free — including the heading ladder — whereas CSS would mean overriding ~15 classes per stylesheet. Levels are logarithmic (`level = log(factor)/log(1.2)`).

Reaching it needs two reflection hops: `WebView.UnderlyingBrowser` (internal getter) then `BaseCefBrowser.UnderlyingBrowser` (protected); `GetHost()` is public from there. **This is the codebase's first reflection into CEF internals**, so `CefBrowserAccess` concentrates it in one place, degrades to a null return on every failure path, probes at startup so a break shows up in the log rather than as a mysteriously dead shortcut, and is pinned by tests that fail the *build* — including on property **type**, not just name — if a package bump breaks the chain. #570 needs the same access for find-in-page.

## Bindings

Four routes, exactly one firing per keystroke: the Windows/Linux window handler, macOS NativeMenu View items, floating windows, and a JS keydown capture for when CEF holds focus.

Keystrokes are recognised by `ZoomKeys`, which matches on **`PhysicalKey` as well as `Key`** — see the review section for why that is not optional here. Every spelling resolves: unshifted `=` (what ⌘+ actually delivers), the shifted form, and the numpad. macOS additionally handles the spellings a NativeMenu item cannot advertise, since it carries one gesture each, while skipping the three the menu owns so nothing fires twice.

View menu gets Zoom In / Zoom Out / Actual Size in both window types. The menu is the discoverability: an undiscoverable feature gets worked around rather than found.

## Ctrl+scroll (#571)

Now routes through the same path on Windows/Linux — stepping the same ladder, persisting, and restoring the reading position. That is #571's fix *by construction* rather than a patch bolted onto Chromium's own zoom.

macOS is excluded: trackpad pinch arrives as ctrl+wheel there, and pinch is page-scale magnification that keeps position and does not rewrap — working behaviour this change was not asked to replace.

## Reading position

Preserved across the reflow using the #434 token, with the resize path's two-part discipline: snapshot **once** on the first step of a burst, restore after the reflow settles. Snapshotting per step would let the 200 ms status tick capture the already-drifted position mid-burst and commit it.

At load time the scroll half of the restoration waits for the **`CACHE_BUILT` signal** rather than a fixed delay, and carries a **navigation generation** so a deferral belonging to a superseded document cannot consume the next one's intents.

## Review

Fable-reviewed twice; seven defects found and fixed, none of which I would have caught alone.

**First pass:**

- **A burst-snapshot bug.** Holding ⌘+ re-snapshotted the reading position on every step, so the 200 ms status tick would capture the already-drifted position mid-burst and commit it — a bounded recurrence of the exact #571 symptom this feature exists to fix. My comment claimed it mirrored the resize path's guard; it had copied only the timer re-arm half.
- **⌘0 was a dead key** whenever the browser's real zoom had drifted from the stored value (a Ctrl+scroll landing before the bridge injects, or a macOS pinch). And CEF's `SetZoomLevel(0)` is the only thing that clears page scale, so the no-op was forfeiting the only way to undo a pinch.
- **A load-time ordering race**: a scroll target could be resolved against the pre-zoom layout, with no rolling token yet to correct it.
- **A macOS gap** where only the unshifted spellings worked outside a book.

**Second pass, on those fixes:**

- **Input-source dependence — the one that matters most here.** Avalonia derives `Key` on macOS from the character the *active input source* produces. With a Devanagari or other Indic layout active, the `=` key doesn't produce `=`, the lookup misses, and the QWERTY fallback reports `OemMinus`. ⌘⇧= would have been silently dead for exactly the readers most likely to have such a layout. Hence `ZoomKeys`, which consults `PhysicalKey` too, on both platforms.
- **A stale deferral across navigations**: a script change inside the deferral window would let the old timer consume the intents the new navigation had just queued, leaving it *worse* off than the race the deferral fixed. Now generation-tagged with an `Interlocked` claim.
- **The 250 ms was an open-loop guess** against a renderer IPC round-trip nothing in this codebase bounds. Now driven by the `CACHE_BUILT` signal — which fires after `fonts.ready` plus a paint frame, i.e. genuinely post-reflow — with the timer demoted to a backstop.

Both passes also **verified**: `SetZoomLevel` self-posts to the CEF UI thread, so calling it from the Avalonia thread is safe (`GetZoomLevel` does **not** — worth knowing for #570); exactly one route fires per keystroke in all five focus/platform combinations; and Chromium's per-origin zoom keys on the full file spec, so a book's zoom cannot leak into the Welcome page, dictionary pane or PDF viewer.

## Not included

The Settings readout and reset from the #572 decision record. It is specified to sit beside the per-script face, and #42 is rebuilding that panel; `IsZoomed`/`FormatZoom` exist for it.

## Testing

Full suite **1177 passed, 0 failed**. The injected bridge script was extracted and syntax-checked with `node --check` — it is a C# verbatim string, so a brace slip there would break the whole JS bridge silently.

**Needs a Windows pass on Merlin**, none of which is testable on macOS:
- that the zoom keys resolve on real hardware, including under a non-Latin input source (the `PhysicalKey` path)
- that `preventDefault` genuinely suppresses Chromium's ctrl+scroll layout zoom in this OSR build — the #571 acceptance case
- the View menu's gesture text, which may render literally as "Ctrl+OemPlus"

Closes #572. Fixes #571.

### DIFF
--- a/src/CST.Avalonia.Tests/Input/ZoomKeysTests.cs
+++ b/src/CST.Avalonia.Tests/Input/ZoomKeysTests.cs
@@ -1,0 +1,152 @@
+using Avalonia.Input;
+using CST.Avalonia.Input;
+using Xunit;
+
+namespace CST.Avalonia.Tests.Input;
+
+// #572: which keystrokes count as book zoom. Matching on Avalonia's Key alone is not safe here, because on
+// macOS Key is derived from the character the ACTIVE INPUT SOURCE produces. With a Devanagari or other
+// Indic layout active — entirely normal for a Pāli reader — the "=" key does not produce '=', the character
+// lookup misses, and Avalonia's QWERTY fallback reports it as OemMinus. Matching Key only would make Cmd+
+// silently dead for exactly the users most likely to be affected. PhysicalKey is layout-independent, so
+// both are consulted. (fable review)
+public class ZoomKeysTests
+{
+    private const KeyModifiers Cmd = KeyModifiers.Meta;
+
+    private static KeyEventArgs Press(Key key, PhysicalKey physical, KeyModifiers modifiers) =>
+        new() { Key = key, PhysicalKey = physical, KeyModifiers = modifiers };
+
+    // ---- The ordinary ASCII-layout cases ----------------------------------------------------------
+
+    [Theory]
+    [InlineData(Key.OemPlus, PhysicalKey.Equal)]            // unshifted "=", what Cmd++ really delivers
+    [InlineData(Key.Add, PhysicalKey.NumPadAdd)]            // numeric keypad
+    public void ZoomIn_Spellings(Key key, PhysicalKey physical)
+    {
+        Assert.Equal(ZoomCommand.In, ZoomKeys.Match(Press(key, physical, Cmd), Cmd));
+        // "+" is Shift+"=" on most layouts, so the shifted form must resolve identically.
+        Assert.Equal(ZoomCommand.In, ZoomKeys.Match(Press(key, physical, Cmd | KeyModifiers.Shift), Cmd));
+    }
+
+    [Theory]
+    [InlineData(Key.OemMinus, PhysicalKey.Minus)]
+    [InlineData(Key.Subtract, PhysicalKey.NumPadSubtract)]
+    public void ZoomOut_Spellings(Key key, PhysicalKey physical)
+    {
+        Assert.Equal(ZoomCommand.Out, ZoomKeys.Match(Press(key, physical, Cmd), Cmd));
+        Assert.Equal(ZoomCommand.Out, ZoomKeys.Match(Press(key, physical, Cmd | KeyModifiers.Shift), Cmd));
+    }
+
+    [Theory]
+    [InlineData(Key.D0, PhysicalKey.Digit0)]
+    [InlineData(Key.NumPad0, PhysicalKey.NumPad0)]
+    public void Reset_Spellings(Key key, PhysicalKey physical)
+    {
+        Assert.Equal(ZoomCommand.Reset, ZoomKeys.Match(Press(key, physical, Cmd), Cmd));
+    }
+
+    // ---- The non-Latin input source case, which is the reason this class exists -------------------
+
+    [Fact]
+    public void IndicLayout_EqualsKeyReportedAsOemMinus_StillZoomsIn()
+    {
+        // Avalonia's QWERTY fallback maps the physical "=" key to OemMinus when the layout produces a
+        // non-ASCII character. Matching on Key alone would zoom OUT here, or (since the shifted form is
+        // unbound) do nothing at all. PhysicalKey.Equal has to win.
+        var e = Press(Key.OemMinus, PhysicalKey.Equal, Cmd | KeyModifiers.Shift);
+        Assert.Equal(ZoomCommand.In, ZoomKeys.Match(e, Cmd));
+    }
+
+    [Fact]
+    public void IndicLayout_RealMinusKey_StillZoomsOut()
+    {
+        // The same fallback reports a genuine "-" as OemMinus too. This must NOT be captured by the
+        // physical-Equal rule above — the two cases are distinguished only by PhysicalKey.
+        var e = Press(Key.OemMinus, PhysicalKey.Minus, Cmd);
+        Assert.Equal(ZoomCommand.Out, ZoomKeys.Match(e, Cmd));
+    }
+
+    [Fact]
+    public void UnknownKeyOnAKnownPhysicalKey_StillResolves()
+    {
+        // A layout we have never seen produces some other character on the "=" key.
+        Assert.Equal(ZoomCommand.In, ZoomKeys.Match(Press(Key.OemQuestion, PhysicalKey.Equal, Cmd), Cmd));
+    }
+
+    // ---- Non-matches ------------------------------------------------------------------------------
+
+    [Fact]
+    public void WithoutTheCommandModifier_NothingMatches()
+    {
+        Assert.Null(ZoomKeys.Match(Press(Key.OemPlus, PhysicalKey.Equal, KeyModifiers.None), Cmd));
+        Assert.Null(ZoomKeys.Match(Press(Key.OemPlus, PhysicalKey.Equal, KeyModifiers.Shift), Cmd));
+    }
+
+    [Fact]
+    public void WithAlt_NothingMatches()
+    {
+        // Keeps this off system and app combos built on Alt.
+        Assert.Null(ZoomKeys.Match(Press(Key.OemPlus, PhysicalKey.Equal, Cmd | KeyModifiers.Alt), Cmd));
+    }
+
+    [Theory]
+    [InlineData(Key.D1, PhysicalKey.Digit1)]
+    [InlineData(Key.O, PhysicalKey.O)]
+    [InlineData(Key.F, PhysicalKey.F)]
+    [InlineData(Key.D, PhysicalKey.D)]
+    public void OtherShortcutKeys_AreNotZoom(Key key, PhysicalKey physical)
+    {
+        // Zoom is matched AFTER the letter-shortcut list in both window handlers, but it must not claim
+        // those keys even if the order ever changed.
+        Assert.Null(ZoomKeys.Match(Press(key, physical, Cmd), Cmd));
+    }
+
+    [Fact]
+    public void TheWindowsCommandModifierWorksToo()
+    {
+        var e = Press(Key.OemPlus, PhysicalKey.Equal, KeyModifiers.Control);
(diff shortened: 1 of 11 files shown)


fsnow:
**Manual verification on macOS (maintainer):** zooming from the smallest rung to the largest and back recovers the reading position every time.

Worth recording precisely what this confirms. On a build taken *before* the second review round, the reading position could be made to drift — but only by zooming **fast**. That is the signature of the burst-snapshot defect: the pre-fix code re-snapshotted on every step, so a rapid burst let the ~200ms status tick capture the already-drifted position between steps and commit it, while a slow zoom always settled before the next step arrived and so never showed it. The `_zoomInProgress` guard takes the snapshot once, on the first step, while the token is still pre-zoom.

That defect lives in view code driven by timers, where a unit test cannot reach it, so this run is the only evidence the fix holds. It also exercises #571's acceptance criterion on macOS — though the Windows Ctrl+scroll path still needs its own pass, since that is a different entry point into the same code.

Still unverified: floating windows, two tabs of the same script following each other, pinch-then-⌘0 clearing page scale, and relaunching with a zoomed script (the load-time deferral path).

Issue #587, opened 2026-08-09
Surface B: evaluation harness — grounding, citation, terminology, marker discipline


fsnow:
Part of #186 (surface **B**). Plan: [`docs/features/planned/AI_SURFACE_B.md`](https://github.com/fsnow/cst/blob/main/docs/features/planned/AI_SURFACE_B.md) §13. The B-side analog of the surface-C cold-agent loop (`docs/testing/LOCAL_API_COLD_TESTS.md`).

A fixed eval set — passages x presets x models — run across the model matrix. **The only surface-B tests that spend money.**

Scored on:

- **Stayed inside the supplied passage** (did not answer from training data).
- **Invented no reference.**
- **Honored the terminology conventions.**
- **Marked quoted Pali correctly** — this is the data that decides whether the v1.1 script conversion is safe to enable, and per model tier.
- **Produced a defensible translation.**

Two things it feeds back into the product:

- **Terminology compliance gates the `recommended` tier** (#584) — a model that cannot follow the convention is arguably not recommended-tier for this app. This is the enforcement point, deliberately instead of post-filtering model output.
- **Marker discipline per model** decides the v1.1 rollout of Pali-script conversion in answers.

`Provenance` on the bundle (#580) stamps app version, corpus revision and dictionary source versions so cross-release regressions are attributable.

**Depends on:** #583. **UI-free.**

fsnow:
## First eval case, observed in the wild — an inverted core term

Recording this while working #578, since it is exactly the shape of finding this harness exists to catch and it would otherwise be lost.

**Setup:** `gpt-oss:120b-cloud` via Ollama's OpenAI-compatible surface, **ungrounded** — no passage, no gloss, answering from memory. Prompt: *"What does the Pali word appamada mean?"*, system prompt *"Answer in one short sentence."*

**Answer:**
> "Appamada" means heedlessness or negligence, especially the lack of mindful vigilance in practice.

**That is the definition of its opposite.** *Appamāda* is heedfulness/diligence — literally *a-pamāda*, non-negligence. The model gave the meaning of *pamāda*, on the most quoted word in the Dhammapada (`appamādo amatapadaṃ`, Dhp 21).

The prose was fluent and confident, and the model spent **56 reasoning deltas** arriving at it — a useful reminder that reasoning volume is not accuracy.

### Why this is a good case for the fixed set

- **The correct answer is authoritative and checkable** — the corpus and DPD both settle it, so scoring needs no judgement call.
[...]

Issue #585, opened 2026-08-09
Surface B: Settings UI — provider, endpoint, model, key, answer language


fsnow:
Part of #186 (surface **B**). Plan: [`docs/features/planned/AI_SURFACE_B.md`](https://github.com/fsnow/cst/blob/main/docs/features/planned/AI_SURFACE_B.md) §7, §9, §10. **UI work — kestrel.**

Extends the existing `AiSettings` (`Models/Settings.cs`), which today holds only the master switch and the `LocalApiSettings` sub-permissions.

- Provider choice, **base URL** (this is what makes DeepSeek/OpenRouter/Ollama/LM Studio work), model name, API key entry.
- **Answer language** and **Pali script in answers** — two separate axes, decided together rather than a release apart.
- The **fidelity advisory** at model-selection time (#584).
- **Do not offer a temperature control** — current Claude models reject sampling parameters with a 400.
- **State plainly what leaves the machine**: passage text, glosses, the user's question, and the system prompt go to the configured provider. The fully local option is real — a local runner through the OpenAI-compatible adapter sends nothing off the machine.

**Depends on:** #579, #584.

fsnow:
**DPD is a prerequisite for surface B** — decided with @fsnow on 2026-08-10 (see [AI_SURFACE_B.md §4](https://github.com/fsnow/cst/blob/main/docs/features/planned/AI_SURFACE_B.md) and #583).

Settings implication: this screen is where a missing DPD should be *surfaced and fixed*, not merely reported. DPD already installs through the existing dictionary-update path, so the natural shape is a prompt to install it alongside the provider/model fields rather than a bare warning — enabling AI features without it produces answers whose gloss half is largely empty.

Two related notes for this issue:

- **Which dictionary feeds glosses** is now partly settled — DPD is preferred where present. Whether Childers 1875 (#378) is a fallback or a supplement is still a data-quality call, and it belongs here.
- **Still no temperature control** (from #578): current Claude models reject sampling parameters with a 400.

Issue #584, opened 2026-08-09
Surface B: model registry and the fidelity advisory


fsnow:
Part of #186 (surface **B**). Plan: [`docs/features/planned/AI_SURFACE_B.md`](https://github.com/fsnow/cst/blob/main/docs/features/planned/AI_SURFACE_B.md) §7, under AI_INTEGRATION.md §11.1's decided policy: **curate, advise, never block.**

- A **model registry** shipped as data: `{ providerId, modelId, tier }` where tier is `recommended` / `permitted` / `discouraged-for-translation`, plus **`unrated`** for anything not listed.
- **`unrated` is not `discouraged`.** New frontier models appear constantly; a registry that punishes anything it has not heard of ages badly and would have flagged half of today's good models a year ago.
- The advisory fires **at selection time in Settings** and as a **badge on translate output**: *this model is not in the recommended tier for translating canonical text; check its output against the Pali.* Never a block.
- Recommended tier defaults **Claude-first**, per the standing policy.

Model quality is a fidelity feature here, not a preference: this is a sacred-text corpus, and hallucination risk is inverse to model quality times grounding.

**Depends on:** nothing. **UI-free** (the data; the surfacing is in the Settings issue).

fsnow:
Tier assignments should cite evidence rather than assert it. The evidence base is now [`docs/testing/PALI_FIDELITY_CASES.md`](https://github.com/fsnow/cst/blob/main/docs/testing/PALI_FIDELITY_CASES.md) — Pāli questions with authoritative answers, tracked as durable cases with dated per-model observations.

First entry is directly relevant here: `gpt-oss:120b` defined *appamāda* as "heedlessness or negligence" — the meaning of *pamāda*, its opposite — fluently and confidently, ungrounded. That is the concrete shape of the risk §11.1 describes when it calls model quality a fidelity feature rather than a preference.

Note the registry design keeps `unrated` distinct from `discouraged`, so a case failure is a reason to rate a model, not a reason to punish an unfamiliar one.

Issue #583, opened 2026-08-09
Surface B: orchestrator — bundle to prompt to provider, streaming and usage accounting


fsnow:
Part of #186 (surface **B**). Plan: [`docs/features/planned/AI_SURFACE_B.md`](https://github.com/fsnow/cst/blob/main/docs/features/planned/AI_SURFACE_B.md) §5, §10.

Ties the pieces together: assemble bundle, render template, call provider, stream deltas to the view model, account for usage.

- **Concurrency is cancel-and-replace, not queue.** Invoking a preset while a stream is running cancels the in-flight one. That is the right semantic for a reader.
- **Cancellation** — a visible stop control; the token cancels the HTTP stream.
- **Usage** — surface tokens in/out and, where the provider reports it, cost after each call. The user is paying.
- **Prompt logging** — bundle contents and responses contain the user's question. Do not log them above Debug.
- **Failure states**: not-configured (explain what to set, link to Settings — never a raw error), offline (say so plainly, do not retry-storm), and the normalized `AiError` set from #578.

### Acceptance

End-to-end with a fake provider that echoes the prompt back, asserting the assembled request. Cancellation and cancel-and-replace covered.

**Depends on:** #578, #579, #582. **UI-free.**

fsnow:
**DPD is a prerequisite for surface B** — decided with @fsnow on 2026-08-10 while building #580. Recorded in [AI_SURFACE_B.md §4](https://github.com/fsnow/cst/blob/main/docs/features/planned/AI_SURFACE_B.md).

It is not an enhancement to grammar answers; it is how glossing works at all. Running Pāli is almost entirely inflected — the text reads `appamādo`, the dictionary holds `appamāda` — so matching surface forms against headwords misses the correct entry for most words. DPD resolves form → lemma and the lemma is what gets glossed.

**What that means for this issue.** The bundler (#580) reports an absent DPD honestly as `BundlePartState.Unavailable` and carries on, because a component should describe reality rather than assume a precondition. **Enforcing the prerequisite is the orchestrator's call**: block the invocation with a clear explanation, or let it proceed degraded with a visible warning. Whichever, it should not be discovered by the user as a mysteriously thin answer.

Also from the same session, and load-bearing for the prompt this orchestrator will send (#582):

- **Nothing injected is authoritative.** Glosses arrive as candidates carrying provenance (`Exact`, `ViaLemma`, `Neighbour`), and a form that is a homograph yields several readings deliberately — the model chooses.
[...]

PR #599, opened 2026-08-11
Surface B: presets, system prompt, and the grounding contract (#582)

### LINKED ISSUE #582: Surface B: presets, system prompt, and the grounding contract
Part of #186 (surface **B**). Plan: [`docs/features/planned/AI_SURFACE_B.md`](https://github.com/fsnow/cst/blob/main/docs/features/planned/AI_SURFACE_B.md) §4, §6.

Each preset is a triple: which bundle pieces to gather, the token budget, the instruction template. Templates are data (embedded resources), user-editable with reset-to-default.

| Preset | Gathers | Requires |
|---|---|---|
| **Explain** | passage, book context, light glosses | — |
| **Translate** | passage, full glosses, apparatus notes | — |
| **Grammar** | selection (or sentence), lemmas, glosses | #581 + lemma asset |
| **Word-by-word** | selection, lemmas + glosses per token | #581 + lemma asset |

Both lemma presets must **degrade visibly** when the optional DPD-lemma asset is absent, never silently.

### The grounding contract — §6

Injection's characteristic failure is not hallucination; it is **truthful output over a silently narrower scope than the user assumed**. "Explain this sutta" against a ~1200-char window produces a confident, correct-looking answer about three paragraphs. Four fences:

1. **Display the scope next to the answer** — rendered by us from `CitationRef`.
2. **Instruct the model to name its scope**, and to say so when the passage does not contain the answer.
3. **Refuse cross-corpus questions explicitly** — "where else does this phrase occur", "who is Ananda" need search or DPPN, not the passage. Direct a graceful refusal pointing at surface C rather than letting the model improvise from training data, which is the invented-citation hazard.
4. **A trimmed passage gets a visible "partial passage" badge**, driven by `BudgetReport`.

### System prompt also carries

- House terminology conventions, stated **positively** ("use 'Pali texts', 'the Tipitaka', 'VRI texts'") rather than as a prohibition — positive phrasing outperforms don't-lists.
- The **Pali-quote marking instruction** from day one (see #587 / §9): v1 strips the markers and renders Latin, but the instruction ships now so the eval harness baselines on the final prompt shape.

**Depends on:** #580. **UI-free.**

fsnow:
Closes #582. Part of epic #186. Plan: [`AI_SURFACE_B.md`](https://github.com/fsnow/cst/blob/main/docs/features/planned/AI_SURFACE_B.md) §4, §6, §9.

#580 decided *what* to gather. This decides *how it is said*, and what the model is told about the difference.

## What's here

- **`Resources/Ai/*.md`** — the shared system prompt and one instruction template per preset, embedded.
- **`PromptTemplateStore`** — built-in defaults, user override in `<app data>/ai-templates/`, validate/save/reset.
- **`PromptBuilder`** — `AiContextBundle` → `RenderedPrompt(System, UserContent, MaxOutputTokens, Notices)`.
- **`PaliQuoteMarkers` / `PaliQuoteFilter`** — the marker contract and its streaming-safe v1 reader.

`#583` composes the `ChatRequest`: this layer deliberately knows nothing about the configured model.

## The grounding contract

The system prompt carries §6's fences, because injection's characteristic failure is not hallucination but a **truthful answer over a silently narrower scope than the user assumed**. It opens by stating what the excerpt actually is, instructs the model to name its scope, refuses cross-corpus questions toward the reader's own search rather than letting it improvise from training data, and states that nothing injected is authoritative and that **absence is not evidence**.

## Two design calls worth a look

**The template engine is substitution-only** — no conditionals, no loops. These are user-editable (#585), and control flow makes them a small programming language whose breakage we then diagnose. The cost is that a placeholder can't be omitted when it has nothing to say, so every value renders as a self-describing sentence instead. That's what makes *"degrade visibly, never silently"* fall out of the design rather than rest on discipline: a missing word-analysis asset says so, in the prompt **and** in a user-facing notice, instead of leaving a blank heading the model reads as an absence in the text.

**A bad override falls back to the built-in, loudly.** Silent fallback would leave the user reading answers from a prompt they believe they replaced — so the reasons are recorded and surfaced as a notice.

## The marker is `[[…]]`

Instructed from v1 though v1 only strips it, so #587 baselines on the final prompt shape. A survey of all 217 books found **zero** occurrences of every candidate (`[[`, `«`, `⟦`, `{`, `|`, `¶`), so corpus collision decided nothing. Prose collision did: `« »` is the standard quotation mark in Russian and French, and a user answering in either would have every ordinary quotation read as marked Pāli. `⟦…⟧` would measure a weak model's Unicode fidelity rather than its instruction-following, which inverts what #587 is trying to learn.

Unbalanced markers are stripped anyway and **counted** — a literal `[[` on screen is a defect for a rule the user never asked about, but the imbalance is signal for #587.

## What the live run found that review didn't

One call to `gpt-oss:120b-cloud` through the #578 provider layer:

- **Reading a dumped prompt** (not a test) caught the inventory claiming `apparatus — given` on presets whose template never rendered it — the model told it had been given something it couldn't see. The inventory now derives from the template's own placeholders, so the two agree by construction, including after a user edit.
- **Every fence held** — cross-corpus refusal, scope naming, no invented references, house terminology.
- **Marker discipline failed on inline terms**: block quotes marked, `*appamāda*` in italics. Fixed by *naming the competing convention* — "an italicised word is a word left behind in Latin" — where stating the requirement had not worked. Verified by a second run.
- **It mistranslated the verse in front of it** — `maccuno padaṃ` as "the samsaric condition", `yathā matā` read from *maññati* rather than *marati*. Not fixable by prompting; filed as **case 2** in `PALI_FIDELITY_CASES.md`, which is exactly the evidence #584's fidelity advisory exists to carry. Case 1's open follow-up also gets its first (incidental) grounded observation.

## Testing

53 new tests; **1257/1257** pass. Every shipped template is asserted against the same validation user edits face — they're data, so nothing about them is checked at compile time.

### DIFF
--- a/docs/features/planned/AI_SURFACE_B.md
+++ b/docs/features/planned/AI_SURFACE_B.md
@@ -1,6 +1,8 @@
 # Surface B — the in-app model, v1 by context injection (Planned)
 
-**Status:** Planned. Not started.
+**Status:** In progress. Shipped: B1 (#578, provider layer), B3 (#580, context bundler), B4 (#582, presets and
+the grounding contract), plus the reader-state read path and `POST /v1/ai/context-preview` (#593). No AI feature
+is user-visible yet — the orchestrator (B5, #583) is what first connects a prompt to a provider.
 **Parent:** [AI_INTEGRATION.md](AI_INTEGRATION.md) — the design of record for the A–E surface map. §11.1 there
 states the decided model-access *policy*; this document is the *implementation plan* for B.
 **Tracker:** epic #186 → children #578–#587 (filed 2026-08-09; the per-item numbers are in §12).
@@ -319,6 +321,27 @@ conversion is safe to enable per model tier.
 v1.1 conversion proviso: **validate that marked content is actually convertible** (pure Pāli character set)
 and leave it Latin otherwise — models will occasionally mark English or emit malformed diacritics.
 
+### The marker is `[[…]]` — decided 2026-08-11 (#582)
+
+Two constraints. It must not occur in the corpus, and it must not occur in ordinary answer prose in **any**
+language a user might set as their answer language.
+
+A survey of all 217 XML books found **zero** occurrences of `[[`, `]]`, `«`, `»`, `⟦`, `⟧`, `{`, `}`, `|`, `~`
+and `¶`; `^` appears 3 times, `` ` `` 6 times, `§` 53 times. So corpus collision eliminates none of the leading
+candidates. (Incidentally: braces being absent from the source confirms that the ones around inline notes are
+added by the passage renderer.)
+
+The **second** constraint is what decides it. `«…»` is out: it is the standard quotation mark in Russian and
+French, so a user answering in either would have every ordinary quotation read as marked Pāli — a v1.1
+conversion bug seeded in v1. `⟦…⟧` is collision-free but measures a weak model's *Unicode fidelity* rather than
+its instruction-following, which inverts what B9 is trying to learn. `[[…]]` is emittable by any model, absent
+from the corpus, and claimed by no language's punctuation.
+
+**Unbalanced markers are stripped anyway and counted, never rendered.** A literal `[[` on screen is a visible
+defect in the answer for a rule the user never asked about; the imbalance is real signal about the model, so it
+is recorded for B9 rather than silently repaired. Marker fragments split across stream deltas are handled by a
+hold-back filter — the same hazard as the think-tag filter, and the same failure if ignored.
+
 ---
 
 ## 10. Failure, privacy, and terminology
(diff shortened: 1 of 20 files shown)


fsnow:
## Fable adversarial review — findings addressed

Full review run against the branch. Everything actionable is fixed in the follow-up commit; nothing was deferred.

### The two that mattered

**H1 — `PaliQuoteFilter` rendered a literal `[[`.** Real, and I traced it: `Strip("[[[[appamāda]]]]")` → `"[[appamāda"`. While inside a quote the loop hunted only for a closer, so a second opener passed through as content — a direct violation of what the class, its tests, and §9 all promise. Not hypothetical either: `[[…]]` is wikitext link syntax, so a model with wiki in its training set will double or nest the markers. Now whichever marker comes first wins in both directions, and the imbalance is counted. Asserted over twelve malformed shapes rather than one.

**H2 — the store is a DI singleton mutating a plain `Dictionary` on every `Get`**, including the no-override path, while `RejectedOverrides` handed out the live map. #583's orchestrator plus a surface-C agent hitting the API is concurrent `Build` by construction — the ordinary cold-agent setup. Now a `ConcurrentDictionary` with a snapshot accessor.

### The validation story was softer than its own documentation

- **M1 — `{{selection}}` and `{{userQuestion}}` are now required.** An override dropping them validated cleanly; the user then selected a phrase, typed a question, and got an answer that saw neither. Worse than the equivalent for the passage, because they *watched themselves* supply the input. This is the one silence the design promises not to permit.
- **M3 — near-miss placeholders.** `{{user_question}}`, `{{2lemmas}}`, `{{ pass age }}`, an unclosed `{{passage` — none match the identifier grammar, so the unknown-name check couldn't see them and they reached the model as literal braces. A residue pass now catches them.
- **M4 — `Reset("../../notes")`** deleted an arbitrary `.md`. Validated like the other two paths.

### The rest

- **M2** — the lemma section claimed "the lookup returned no candidates" for presets that run no lookup. Fixed, and reordered so candidates in hand are always reported as candidates whatever the budget report says.
- **M5/M6/L6** — `translate.md` said to give the Pāli "in brackets", colliding with the marker syntax the same prompt had just taught (and inline marking is exactly where the live run showed discipline failing). The terminology bullet had no antecedent for "them"; it now names its subject and asks for the terms *in whatever language the answer is written in*.
- **L1** — the two templates were scanned concatenated, which could synthesize a placeholder at the seam. Separator added. Now unreachable anyway, since M3 rejects a stray `{{`.
- **L2** — the budget comment's arithmetic was wrong (Explain is 1600 chars vs word-by-word's 600, not 4×), and the reviewer is right that 3000 may not be generous at the boundary. Corrected, and flagged in the comment as reasoning rather than measurement for #587 to check.
- **L3** — the scope statement told the model a window covering a short book was "not the whole text", buying hedging it didn't need.
- **L5** — markdown escaping applied only to the gloss; stem and part-of-speech come from the separately versioned DPD asset, which this code shouldn't assume stays pipe-free.
- **L4** — §6's fence 4 promised a partial-passage badge that #580 removed the driver for. Recorded in the doc as *delivered differently*, with why, rather than left for the next reader to file as a gap.

**Not taken:** the size cap on `Save` — the user is editing their own file, and a cap would be one more thing to explain. The eager `BuildValues` is left alone; the `shown` set could gate it, but that trades clarity for work that is already negligible.

Re-verified against `gpt-oss:120b-cloud` after the prompt edits: markers, cross-corpus refusal, scope naming and terminology all still hold. **1270 tests pass.**

Issue #573, opened 2026-08-07
Configured font missing from the system is substituted silently (and Settings shows it as "System Default")


fsnow:
When a font that the user has configured is no longer installed, the app **substitutes silently**. Nothing in the UI says the chosen font is gone, and Settings actively shows a value the user did not choose.

This is not hypothetical. A preferred Devanagari font disappeared from macOS in 26.5, and the only symptom was that the text started looking wrong.

## Why silent substitution is worse here than in a typical app

Faces vary enormously in apparent size at the same point size, so a user picks a face and then adjusts until the text looks right — via zoom (#572), which is stored per script. That adjustment is **calibration for a specific face**, not a general size preference.

So when the face silently vanishes, the surviving zoom level is calibrated for **nothing**. The substitute renders scaled by a factor chosen for a different font's metrics, and the result is text that is simply the wrong size with no visible explanation. The user is left to guess whether they mis-set something, whether the app regressed, or whether the OS changed underneath them.

There is no way to predict which fonts will be present on a user's machine, so this will keep happening.

## Current behaviour

The **data** is safe already — #67 fixed that. A missing font does *not* erase the saved preference:

```
SettingsViewModel.cs:485  // Its display value is decoupled from the saved FontFamily so a load can
                          // show "System Default" for a temporarily-missing font WITHOUT erasing the
                          // saved value. Only a genuine user pick persists. (#67)
```

So reinstalling the font restores the setting. Good.

What is missing is the **notification**. Concretely:

- `FontService.GetScriptFontFamily` (`FontService.cs:41`) returns the configured name with **no check that it resolves**. Callers hand it to Avalonia / the WebView, which quietly falls back.
- `GetAvailableFontsForScriptAsync` (`FontService.cs:122`) already enumerates exactly what is installed — the data needed for the check is on hand and pre-cached at startup by `PreloadFontsForAllScriptsAsync`.
- Settings shows **"System Default"** for the missing font. That is a display-only placeholder protecting the saved value, but the user cannot tell it apart from having genuinely selected System Default. The one screen that could explain the situation instead disguises it.

So every ingredient for detection exists; nothing consumes it.

## What to build

1. **Detect** — on startup and on settings load, compare each configured script font against the available-fonts list already cached for that script. Same check when book faces become configurable (#42).
2. **Report** — tell the user plainly, naming the missing font. Something with the shape of:
[...]

PR #612, opened 2026-08-12
Book fonts: one stylesheet, the face as a per-script setting (#42)

### LINKED ISSUE #42: Book-content font face management (user-configurable font faces, injected at render time)

fsnow:
Book text was rendered from **fourteen near-identical stylesheets** and could not be configured at all. #574 made those fourteen byte-identical apart from their single `font-family` line — which is exactly the precondition this collapses.

## One stylesheet

`tipitaka.xsl`, with the face supplied as an XSLT parameter at transform time.

**Verified rather than assumed:** a real book transformed through each old per-script stylesheet and through the new parameterised one produces **byte-identical output for all fourteen scripts** — including Latin, whose stack contains quoted family names, so `xsl:value-of` isn't over-escaping.

## The face becomes a setting

The fourteen shipped stacks move into `BookFontDefaults` as data. `BookFontResolver` picks the user's choice or the shipped default, per script.

An unset preference stores **empty** rather than a copy of the default, so a user who never chose keeps tracking the shipped stack instead of being frozen against whatever it was the day they first ran the app.

A user-chosen face is **CSS-quoted and escaped**. Not cosmetic: macOS ships "Bodoni 72" and the picker offers it, but `font-family: Bodoni 72;` is invalid CSS — the whole declaration is dropped and the book silently renders in the browser default with nothing to explain it. Control characters and lone surrogates are stripped too, since the `XmlWriter` behind `XslCompiledTransform` rejects them and would fail the entire transform.

## The user-editable XSL directory is retired

The app renders from the stylesheet in its own bundle now. That directory was CST4 heritage, and both reasons anyone edited it are settings today: the **face** (here) and the **text size**, which per-script zoom covers (#572).

It was also actively harmful. The copier only acted on an **empty** directory, so every existing install was still rendering with whatever shipped the day it was first launched — on the maintainer's machine, **13 of 14 files predating #574**, still carrying the old per-script sizes and the seven different `.hangnum` margins. #574's changes had reached nobody, and behaviour silently depended on install age.

The beta 6 migration **moves it aside** rather than deleting. A CST4-style user may have hand-edited those files; making their work inert is a different act from destroying the only copy of it.

## Settings

A **"Pali Fonts (Books)"** group beside a renamed **"Pali Fonts (UI)"** — face only, since size is zoom. Its description carries the **live per-script zoom**, which is also the only place in the app that reveals zoom is stored per script.

The old description claiming the UI section covered books is corrected — it never did.

Also: the script list is alphabetical and full-height with all fourteen visible, opens on the script the app is **currently displaying**, the nav selection is actually visible (it was styled with a brush too faint to see in light mode), both font combos align, and the dialog's **size persists** — applied before showing, since `CenterOwner` positions using the size the window has when shown.

A face change **re-renders open books**, carrying the #434 reading-position token, since the face is baked in at transform time and cannot update in place.

## Review

Fable-reviewed twice; every finding fixed. Three worth recording:

- **The saved face showed a blank ComboBox.** Reproduced in testing *after* I had deferred it as a cosmetic edge case about uninstalled fonts. It blanked for **installed** fonts too, on any case or whitespace difference, because a ComboBox only renders a `SelectedItem` equal to an item of its `ItemsSource`. The chrome picker had always resolved against the list; the book one didn't.
- **The re-render guard recorded the face before the render succeeded**, so a failed reload left the view believing it already showed that face and no later event would retry — the book stuck on an error page until a *different* face was picked. Now a separate in-flight field, cleared on every exit.
- **The migration's already-exists branch deleted on an unprovable assumption.** Both directories existing doesn't prove the current one is stale: a still-installed older build recreates it, so after a state restore the current one can hold the **newer** edits. It picks a free name and never deletes.

## Also

The `CST.Core` converters' stylesheet-PI rewrite now names `tipitaka.xsl` rather than a per-script file that no longer exists. (Cst4's copy legitimately differed — it shipped its own set — and that tree has since been removed from main in #611.)

## Not included

Both specified for this panel in #42, and worth an explicit decision rather than quiet omission:

- the zoom **reset button** — the size line points at ⌘0 instead
- **#573's missing-font reporting** — `BookFontResolver.HasUserChoice` is the hook it will need, since only an explicit choice can go missing

## Testing

Full suite **1454 passed, 0 failed**, including 31 new tests. Among them an end-to-end one that pushes every hostile stored value through a real `XslCompiledTransform` rather than reasoning about the escaping, and three that pin the migration's *move* semantics — the previous set asserted only `!Directory.Exists`, which a delete satisfies equally, leaving the whole fix unpinned.

**Windows verification wanted:** the migration's `Directory.Move` with an open handle in the directory, and one packaged-build smoke test that books render from `xsl/tipitaka.xsl` beside the executable.

Closes #42.


fsnow:
Reviewed independently plus a third adversarial pass. **Merging.**

**Claims verified rather than taken on trust.** Full suite 1454/1454 as stated (now 1455 with the test added below). The 14→1 collapse is complete and accounted for — `tipitaka-beng.xsl` was *renamed* to `tipitaka.xsl`, which is why the diff shows 13 deletions rather than 14 — and all fourteen `BookFontDefaults` stacks byte-match the `font-family` lines of the stylesheets they replace. Deployment holds on all three paths the locator handles: `package-macos.sh` copies `xsl/*.xsl` by wildcard, `package-windows.ps1` stages the whole directory, and nothing anywhere named a per-script file. Repo hard rules clean — no forbidden word in the diff, every font name in `BookFontDefaults` is ASCII, no WebView re-parenting touched. `SettingsWindowState` is backwards compatible: new property with defaults, `??= new()` in the validator, NaN/Infinity sanitised.

**Two findings fixed on the branch in c9df749**, both in the seams around the earlier fixes rather than in the fixes themselves:

- **A leaked in-flight claim on two of four render exits.** `_requestedBookFontFamily` is cleared on the stylesheet-missing branch, the catch and success — but not on `booksDir == null` or `!File.Exists(xmlPath)`. The field's comment says "either way", so the code did not hold its stated invariant. With a misconfigured XML directory, choosing face B leaves the claim set forever and every later event for B is suppressed. This is the defect the two-field design closes, surviving on the exits nobody re-read. Fixed by inspection; pinning it needs the refactor recorded in #613.
- **An empty `font-family` from a value that sanitises to nothing.** Empty families were filtered *before* quoting, but `QuoteFamily` is what drops control characters — so a stored `U+0001` passed the blank guard, came back as `""`, and produced `font-family: "";`: valid CSS matching no family, i.e. tofu in a Devanagari book, which is exactly what the degenerate-value fallback exists to prevent. Now filtered after quoting, with a mutation-verified test.

**Filed, not fixed:** #613 (reverting to the current face mid-render leaves the superseded one displayed — needs a generation counter or a post-transform re-check, not a wider field set), #614 (a saved-but-uninstalled face cannot be cleared; the picker says "Default" while the renderer still uses the missing name), #615 (`char.IsSurrogate` strips valid pairs), #616 (a Beta 5 relaunch recreates `xsl/` and the recorded id then skips the retirement forever).

**Two notes for the record.** The PR body doesn't mention that **Cyrillic rendering changes**: the old `GetXslPath` switch had no `Script.Cyrillic` case, so Cyrillic books fell through `_ => "latn"` and `tipitaka-cyrl.xsl` was dead code. They now get their own stack. It's an improvement and it's documented in a code comment, but it is the one thing here that changes what an existing reader sees. Separately, `ApplicationState.cs` has the new class inserted between `MainWindowState`'s doc comment and its declaration, so `SettingsWindowState` carries two `<summary>` elements and `MainWindowState` has none — cosmetic, one-line fix whenever that file is next touched.

The earlier two Fable passes held up well: the escaping core, the stack byte-match, and the migration's containment all came back clean under adversarial probing.

Issue #541, opened 2026-07-31
Status bar shows every page-numbering system for every book, including ones the book does not have


fsnow:
Part of the #422 cluster. Goal: **displayed numbering systems should be correct and relevant.** Today they are neither — every system is shown for every book.

## The defect

`BookDisplayViewModel.UpdatePageReferencesText` composes the status bar unconditionally:

```csharp
// Temporary hardcoded format until localization is implemented - now includes paragraph number for debugging
var newText = $"VRI: {VriPage}   Myanmar: {MyanmarPage}   PTS: {PtsPage}   Thai: {ThaiPage}   Other: {OtherPage}   Para: {CurrentParagraph}";
```

All five systems appear for every book. A book with no PTS pagination still shows a `PTS:` field — permanently blank or `*` — which reads as missing data rather than as *not applicable*.

That is misleading in a way that matters for a scholarly reader: an empty `PTS:` invites the conclusion that the PTS page is unknown here, when in fact this text has no PTS pagination at all. Missing V/M/P/T numbering is **expected** for large parts of the corpus (the VRI print set is 140 volumes; the extra-canonical texts were not published outside Myanmar) and is not a data gap.

## Also: a debugging field shipped

`Para: {CurrentParagraph}` is described in its own comment as "for debugging". It is in the shipped status bar.

## What it needs

Per-book metadata for **which numbering systems this book actually carries** — the same class of missing model as the absent volume metadata noted on #422, and the same gap that makes #457 derive availability from scroll position.

Once that exists, the status bar renders only the applicable systems, and #457's Go To greying-out reads from the same source instead of guessing.

## Relationship to #457
Inverse symptoms, one root cause:

| | Symptom |
|---|---|
| **#457** | Go To **hides** systems the book *does* have (greys out by scroll position) |
| **this** | The status bar **shows** systems the book *does not* have |

Both are "we do not know what numbering a book carries". Worth fixing from one source of truth rather than separately.

## Note
The hardcoded format is also flagged in-code as temporary pending localization (#26), so whatever replaces it should not re-hardcode the label set.

PR #629, opened 2026-08-13
Book information: an "i" panel naming the source file (#628)

### LINKED ISSUE #628: Feature request: display XML filename in the book information or tab
**Description:**

When working on text corrections, I need to know the exact XML filename of the book I am reading (e.g., s0203m.mul.xml). This is essential for cross-referencing with other tools and for creating issues on GitHub.

Currently, CST Reader does not display the XML filename anywhere in the interface (neither in the tab, nor in the book info panel, nor in the title bar). I have to use another program to look up the filename, which is time-consuming.

It would be very helpful if the XML filename could be shown somewhere.

This feature would greatly improve the workflow for those of us who do text corrections and need to reference the source files directly.

Thank you for considering this.

fsnow:
An error found while reading has to be reported against the **source file**, and the reader was the one tool in that workflow that never showed which file it was displaying — so the name had to be looked up in another program every time.

An **"i"** button in the book toolbar opens a flyout:

```
XML file   s0203m.mul.xml  [Copy]
Book       Vinayapiṭaka / Pārājikapāḷi
Piṭaka     Vinaya
Level      Mūla (root text)
```

## Why a panel rather than another status-bar field

The file name is the ask, but it is not the only thing that identifies a book to someone filing a report — and the status bar already carries five page-numbering systems for every book, whether or not the book has them (#541). A flyout costs one button of toolbar width and leaves room to add to it later.

## Copyable, deliberately

Every value is a `SelectableTextBlock`: the whole point of showing these is to paste them somewhere else. The file name additionally gets a one-click **Copy**, since it is the field that gets pasted every time.

## The one rule worth stating in code

**The file name is never script-converted.** Every other string in this panel is, and a converted file name would look like an answer while matching nothing on disk, in the repository, or in an issue report. The nav path *is* converted, and follows the script the book is being read in — a path shown in the wrong script is worse than no path at all to someone cross-referencing.

## Testing

The formatting is static and pure, so it is tested without a `Book`, a dock factory or a WebView — the same shape as the zoom readout and the book-face decisions. Covered: the conversion, the Devanagari pass-through (a Deva→Deva round trip is a no-op at best, and this panel is where a corrupted path would look authoritative), the empty-path case, and every enum label.

Full suite **1553 passed, 0 failed**, including 16 new tests. Verified in the running app, and a screenshot is on the issue.

Closes #628.

### DIFF
--- a/src/CST.Avalonia.Tests/ViewModels/BookInformationTests.cs
+++ b/src/CST.Avalonia.Tests/ViewModels/BookInformationTests.cs
@@ -27,13 +27,20 @@ public class BookInformationTests
     // ---- The file name ---------------------------------------------------------------------------
 
     [Fact]
-    public void The_file_name_is_shown_exactly_as_it_is_on_disk()
+    public void Putting_a_file_name_through_the_panels_conversion_would_corrupt_it()
     {
-        // It names a file. Converting it — as every other string in this panel is converted — would
-        // produce something that looks like an answer and matches nothing on the filesystem, in the
-        // repository, or in an issue report.
-        var book = new Book { FileName = "s0203m.mul.xml", LongNavPath = DevaNavPath };
-        Assert.Equal("s0203m.mul.xml", book.FileName);
+        // Why XmlFileName must never be converted, demonstrated rather than asserted. Every other string in
+        // this panel goes through FormatNavPath; a file name put through the same call comes back altered,
+        // and an altered file name is worse than none — it looks like an answer while matching nothing on
+        // disk, in the repository, or in an issue report.
+        //
+        // The invariant itself lives at the property (BookDisplayViewModel.XmlFileName returns
+        // Book.FileName untouched) and CANNOT be reached from here: constructing a BookDisplayViewModel
+        // needs ReactiveUI initialized, which the headless test host does not provide. So this pins the
+        // STAKES, not the rule. The rule rests on the property's own doc comment and on review.
+        var converted = BookDisplayViewModel.FormatNavPath("s0203m.mul.xml", Script.Latin);
+
+        Assert.NotEqual("s0203m.mul.xml", converted);
     }
 
     // ---- The nav path ----------------------------------------------------------------------------
--- a/src/CST.Avalonia.Tests/ViewModels/BookInformationTests.cs
+++ b/src/CST.Avalonia.Tests/ViewModels/BookInformationTests.cs
@@ -1,0 +1,110 @@
+using CST;
+using CST.Avalonia.ViewModels;
+using CST.Conversion;
+using Xunit;
+
+namespace CST.Avalonia.Tests.ViewModels;
+
+/// <summary>
+/// The book-information panel's formatting (#628).
+///
+/// <para>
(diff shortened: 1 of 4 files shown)


fsnow:
Reviewed independently plus a Fable adversarial pass. **Two findings fixed on the branch in `b2a91ed`; merging.**

**Verified rather than taken on trust.** Based on current `main` — no rebase needed despite #622 having touched two of the same files. **1553/1553** exactly as claimed. No forbidden word in the diff. The `\uXXXX` rule is not engaged: the diacritics here (`Mūla`, `Aṭṭhakathā`, `Ṭīkā`) are visible Latin-block characters matching existing practice, not the invisible joiners that rule exists for. All new non-ASCII string literals are **NFC** — worth checking in this repo specifically, since `ā` as `a`+U+0304 versus U+0101 renders identically and compares differently.

**Fixed: the headline invariant had no biting test.** `The_file_name_is_shown_exactly_as_it_is_on_disk` asserted `book.FileName` against the literal it had just assigned — a round trip through `Book`'s own property that never reaches `BookDisplayViewModel.XmlFileName`. Changing that property to convert left all sixteen tests green, i.e. the regression the PR names as *"worse than nothing"* was unguarded.

The obvious repair does not work: constructing a `BookDisplayViewModel` throws, because ReactiveUI is not initialized in the headless test host. So the invariant is genuinely unreachable from a unit test. The test now demonstrates the **stakes** instead — a file name put through the same `FormatNavPath` the nav path gets comes back altered, since that call title-cases — and says plainly that the rule rests on the property's doc comment and on review. Claiming coverage that does not exist is worse than admitting the gap.

**Fixed: the nav path refreshed only after a CEF round trip.** It sat below an awaited `GetCurrentPositionTokenAsync`, sharing one try/catch with it, while the font properties refreshed immediately above. A slow JS call left an open flyout showing the **old script's text in the new script's font**; a throwing one left it that way permanently. It reads only `_bookScript`, already updated by the time the subscription runs, so it had no reason to wait on the WebView — moved up beside the font raises.

**Verified clean, including empirically.** Fable ran a real corpus nav path through all fourteen scripts (separators and parens survive each; no empty outputs), confirmed `AvailableScripts` excludes `Unknown`/`Ipe` so the converter's empty-return path is unreachable, and built a headless Avalonia repro to confirm `FindControl` resolves an `x:Name` inside a `Flyout` — the Copy wiring genuinely attaches, which was its strongest suspicion. Enum coverage is complete: 4/4 `Pitaka`, 4/4 `CommentaryLevel`.

**Two nits left alone**, both defensible as-is: the Copy button's failure is logged but invisible to the user, and the Devanagari pass-through test pins the behaviour rather than the mechanism (`ScriptConverter.Convert` short-circuits identity conversions, so deleting the explicit branch still passes).

**One thing I could not resolve.** The first full run after my edits reported a single failure; three subsequent runs of the same binaries — including one captured to a file — were clean at 1553/1553. I never captured which test it was, so I can only say it is not deterministic and does not reproduce from these changes. Flagging it rather than leaving it in the log unexplained.

Issue #633, opened 2026-08-14
Clicking into a book does not reliably give its browser keyboard focus: window activation restores focus to another document view


fsnow:
Clicking into a book does not reliably leave that book's browser with keyboard focus. A single click often works; when it does not, a second click or a double click always does. Found while testing #621 in a three-way split inside a floating window, but it is not specific to splits or to floating windows — it is specific to **activating a window by clicking into a browser**.

## Why it matters more than it looks

When focus is inside a book's WebView, CEF consumes the keystroke and the page relays it to **its own view** (`CST_GOTO`, `CST_CLOSE_TAB`, `CST_ZOOM_*` in `BookDisplayView`). No layout resolution happens — the relay is tagged with that view's tab id, which is correct by construction.

So for every shortcut pressed with focus in a book, **the target is decided entirely by which browser holds keyboard focus**. If focus is on the wrong browser, the wrong book is acted on, and nothing downstream can correct it. #621's document-targeting fix does not apply here, because the keystroke never reaches Avalonia at all.

The worst case is cross-window: if the remembered focus belongs to a book in the **main** window, activating a floating window can deliver the keystroke to a book that is not even visible. Observed as "Go To opened for a book in the other window".

## What the log shows

Activating a window makes Avalonia focus a document VIEW — `source=BookDisplayView`, naming whichever document held focus last, not the one just clicked. That lands within milliseconds of CEF's own focus callback for the browser actually clicked, and **whichever arrives last takes the keyboard**:

```
16:56:53.521  Avalonia focus -> BookDisplayView (s0103)
16:56:54.823  GO TO REQUESTED FROM JAVASCRIPT -> s0103      <- Avalonia's restore won

16:57:15.104  Avalonia focus -> BookDisplayView (s0103)
16:57:16.203  browser focus  -> s0101                        <- CEF's report landed after
16:57:16.905  GO TO REQUESTED FROM JAVASCRIPT -> s0101      <- and won
```

A second click produces no activation restore, which is why it always works.

## Not the same as #621

#621 fixed the **menu/window-handler** path, where a command resolves a target from the layout. That path is now correct and tested. This issue is upstream of it: real keyboard focus landing on the wrong browser, which the relay path then faithfully honours. Both symptoms look identical to a user — "it acted on the wrong book" — which is why they were conflated at first.

## Direction, and the hazard

[...]

Issue #518, opened 2026-07-27
Windows: keyboard shortcuts are dead when focus is in the Welcome, Dictionary, or PDF WebViews


fsnow:
> ### 🪟 Windows only
> This does **not** affect macOS, and macOS cannot reproduce it — see [Why macOS is immune](#why-macos-is-immune). Reproduce on **Merlin** (ARM64 VM) or **Placid**.

## Summary

**On Windows,** keyboard shortcuts are dead whenever focus is inside the **Welcome page**, the **Dictionary meaning pane**, or the **PDF viewer**. Only the book WebView works.

Most visible at startup: the Welcome page holds focus when the app opens, so pressing Ctrl+O or Ctrl+, immediately after launch does nothing.

## Why

Established while fixing #509 and its follow-up: the rule is not *which view* but **whether a WebView holds focus**.

- Focus in Avalonia's own controls (tab strip, tool panes, book tree) → the window `KeyBindings` on `SimpleTabbedWindow` fire normally.
- Focus in a WebView → **CEF takes the keystroke before Avalonia sees it**, so a shortcut only works if that WebView's page forwards it back out via JavaScript.

`BookDisplayView` injects a `keydown` capture that forwards Ctrl/Cmd + C, A, E, ⇧E, G, W, D, F, O, P, ⇧P and `,` back to C# through `document.title`. **The other three WebViews have no such capture at all**, so every shortcut is inert while focus is in them.

Affected views:

| View | Keyboard capture |
|---|---|
| `BookDisplayView` | yes |
| `WelcomeView` | **none** |
| `DictionaryPanel` (meaning pane) | **none** |
| `PdfDisplayView` | **none** |

This is most noticeable on the Welcome page, since it is the tab that has focus at startup.

## Repro

1. Click the Welcome page (or a dictionary meaning, or a PDF)
2. Press Ctrl+, (Settings), Ctrl+O (Select a Book), or Ctrl+D
3. Nothing happens. The same keys work with the tab strip or a book focused.

### Why macOS is immune

`SimpleTabbedWindow.RegisterMenuShortcutKeyBindings` opens with `if (OperatingSystem.IsMacOS()) return;` (#509). On macOS the **system** menu bar registers these gestures as real accelerators, so AppKit consumes the keystroke before CEF is offered it — where focus sits is irrelevant, and the JS forwards are not load-bearing.

[...]

Issue #443, opened 2026-07-21
Menu shortcuts (⌘W/⌘G/⌘E) target the first DocumentDock, not the focused split


fsnow:
In a split document layout (two document panes side by side), the native-menu shortcut handlers resolve their target through `FindDocumentDockInLayout`, which returns the **first** `DocumentDock` found walking the layout tree — not the one the user is actually working in. Its `ActiveDockable` is then used as the command target.

So with focus in the second split, a menu-path shortcut acts on the first split's active tab:

- ⌘W (`OnCloseTabClick`, `SimpleTabbedWindow.cs`) can close the wrong pane's tab;
- ⌘G (Go To) and ⌘E / ⇧⌘E (View Source) have the same resolution and target the wrong book.

The keystroke-with-book-focus route is unaffected — the JS forward carries a tab id and acts on the book that has focus. Only the menu/accelerator route resolves by layout position.

**Suggested fix:** resolve the target from focus rather than tree order — `rootDock.FocusedDockable`, falling back to the current first-`DocumentDock` behaviour when nothing is focused. One shared helper for all these handlers, since they all repeat the same lookup.

Pre-existing; predates #110, which mirrors the existing ⌘G / ⌘E resolution rather than widening the problem. Noted during review of PR #442.

Issue #458, opened 2026-07-21
Crash (SIGSEGV): closing a tab in a multi-book floating window re-attaches a sibling's live CEF browser


fsnow:
Closing a tab in a **floating window that holds more than one book** can SIGSEGV. Reproduced once on 2026-07-21 via ⌘W; the close path is shared with the tab's close button, so this is not specific to the keyboard shortcut.

## Crash

`~/Library/Logs/DiagnosticReports/CST.Avalonia-2026-07-21-162631.ips`

```
EXC_BAD_ACCESS (SIGSEGV), KERN_INVALID_ADDRESS at 0x0
faulting thread: CrBrowserMain (com.apple.main-thread)

libAvaloniaNative  AvnNativeControlHostTopLevelAttachment::InitializeWithChildHandle(void*)
libcef.dylib       ...
```

## Sequence from the log

```
16:26:27.700  Removed book window state ... Book_2_s0103m   <- close of the active tab
16:26:27.700  ◀◀◀ DETACHED  s0103m
16:26:27.705  ▶▶▶ ATTACHED  s0102m  "attached to window for FIRST TIME"
              SIGSEGV
```

Closing the active tab makes Dock activate a sibling, and attaching that sibling's view runs `InitializeWithChildHandle` against a **live CEF browser** — the re-parent that `docs/architecture/DOCK_SUBSYSTEM.md` says must never happen.

Note that plain tab *switching* inside the same floating window survived twice moments earlier (`16:26:22`, `16:26:24`: detach + attach, no crash). Only close-then-activate-sibling crashed, so the trigger is narrower than "any attach".

## Notes toward a fix

The documented discipline is dispose-before-move: evict the recycled view so a fresh browser is built at the destination (`DisposeAndEvictRecycledView`, as the float/unfloat button paths do). Closing a tab currently disposes only the **closed** book's view — the sibling that is about to be re-attached keeps its live browser.

Worth checking whether the same happens in the main window with a multi-tab split, and whether the sibling attach can be deferred past the close (dispatcher post) so the two never overlap.

Related: #39 (intermittent CEF re-parent crash during dock drag) is the same failure class but a different trigger — drag rather than close. Distinct enough to track separately; close as a duplicate if a single fix covers both.

fsnow:
**Confirmed reproducible.** Exact steps (reproduced twice):

1. Open books **B** and **C**.
2. **Float both** into their own windows.
3. **Drag C onto B**, so both books live in a single floating window.
[...]

PR #634, opened 2026-08-14
Target window commands by interaction, not by tree order (#621)

### LINKED ISSUE #621: Zoom keys act on the wrong document: targeting falls back to the first dock whenever focus is not inside a book's WebView
Reported while testing #618, but **not caused by it** — the same thing happens with the ⌘/Ctrl +/− keys alone, which shipped in #572.

## What happens

With a book and a source PDF open side by side in a split, click into the **PDF** — including clicking its own zoom controls — and press ⌘/Ctrl **+** or **−**. The **book** zooms. There is no way to zoom the PDF from the keyboard, and the keystroke silently acts on a document the user is not working in.

Confirmed in the log rather than inferred. A session with the PDF dragged into its own split:

```
22:25:06.135  MoveDockable — MainDocumentDock now holds only the PDF
              (the book's DocumentDock lives inside MainDock, earlier in tree order)
22:25:09.766  FOCUS: SimpleTabbedWindow LostFocus. Source: DocumentTabStripItem
22:25:18.483  Zoom Out from window: CST Reader
22:25:18.483  Book zoom for "Latin" set to 90 %
```

Nine seconds after the window lost focus to the PDF, a zoom keystroke changed the **book's** zoom.

## Why

`DocumentTargetResolver.ResolveActiveDocument` (`src/CST.Avalonia/Services/DocumentTargetResolver.cs`) resolves in three steps: the caller's preferred dockable (derived from Avalonia keyboard focus), else `layout.FocusedDockable`, else **the first document dock in tree order**.

When focus sits inside a CEF WebView, Avalonia has no usable focus to report — the browser owns a native surface, and `RootDock.FocusedDockable` is never populated in this app (the resolver's own comment records that). So resolution falls through to the last step, and `FindFirstDocumentDock` walks depth-first: in the layout above it reaches the **book's** dock before the one holding the PDF. The book is returned, `InvokeZoom` casts it to `BookDisplayViewModel` successfully, and zooms it.

#443 fixed exactly this class of bug for splits, but only for focus Avalonia can see. Focus inside a browser is invisible to it, and a CEF document (PDF, Dictionary, Welcome) is precisely where it lands.

On macOS the keystroke is not swallowed by the PDF because the View menu's key equivalent is claimed by the application menu before the focused view ever sees it — so the command runs regardless of what has focus, and then picks the wrong target.

## Scope

Every window-level command that routes through this resolver shares the flaw: ⌘W (close tab), ⌘G (Go To), ⌘E / ⇧⌘E (View Source), ⌘D (dictionary lookup), ⌘F (find in page). Zoom is simply the most visible, because it changes something the user can see happening to the wrong pane. **⌘W is the one worth checking first** — closing the wrong tab is destructive in a way zoom is not.

Related but distinct: #518 (shortcuts dead *inside* the PDF because Chromium's plugin frame eats them, and `ExecuteScript` cannot reach it). That issue is about keystrokes not arriving; this one is about a keystroke that arrives and is aimed at the wrong document.

## Two parts

**1. Targeting.** A command should act on the document the user is actually working in, even when that document is a browser. Avalonia focus cannot answer this, so something else has to: the most recently *activated* dockable is the obvious candidate, since the dock factory already sees activation when a tab is selected or a view is loaded. It needs care around staleness — a closed tab must not remain the target — and a deliberate answer for "the target is not a book", which today is silently the wrong book instead of nothing.

**2. Zooming the PDF.** Once the PDF is correctly identified as the target, ⌘+/− should zoom it. Mechanically this is the same `CefBrowserAccess.TryGetBrowserHost` → `SetZoomLevel` path the book already uses, so it is small once part 1 exists. Worth deciding whether PDF zoom persists at all: book zoom is stored per script because it calibrates a font face, and neither reason applies to a scanned page image, so ephemeral per-tab zoom is probably right.

Part 1 alone fixes the bug (the keystroke stops acting on the wrong document). Part 2 is what the user was trying to do in the first place.

fsnow:
Window-level commands — zoom, ⌘W, ⌘G, ⌘E, ⌘D, ⌘F — resolve their target from Avalonia's keyboard focus, falling back to *the first document dock in tree order*. A CEF WebView hosts a native surface, so a click into a book, a PDF or the Welcome page leaves Avalonia with nothing to report and the fallback runs. Those three are most of the reader's surface, so **#443's bug was reachable again the moment anyone clicked into a document rather than onto its tab**.

`ActiveDocumentTracker` supplies the signal Avalonia cannot: a bounded, weakly-held history of interactions, read as one new tier between live focus and the first-dock guess.

| Feed | Sees |
|---|---|
| `dock-activation` | Dock's `ActiveDockable` setter — one subscription covering every dock in every window, including ones splits and floating windows create later |
| `browser-focus` | CEF's own focus callback — the only signal that can see a click into a browser body |
| `avalonia-focus` | window-level `GotFocus` — tab-strip clicks (no activation fires when the tab is already active in its own split), toolbar controls, tool panes |

**Everything is recorded, tools included, and filtering happens on read** by asking the caller's layout what it contains. That single rule buys three things: a new document whose `Owner` isn't set yet still registers; a tool activation can't displace the document history; and one flat list serves every window. Keying per window is a trap — Dock's `FindRoot` returns the *nearest* root, the inner `WindowLayout`, while every caller holds the outer `Root`, so the keys would never match.

Live focus still beats history, which is what makes #443 unable to regress. Every pre-existing resolver test passes unchanged, and they double as the guarantee: they pass no history, which is also the "tracker unavailable" production path.

## Two things measured rather than assumed

Both changed the design, and both are the kind of assumption that would have failed silently.

**CEF's `OnGotFocus` does fire for the PDF's internal plugin frame** — the place where the in-page keydown relay is already known to be inert (#518). Had it not, the browser feed couldn't have covered PDFs at all. The same probe settled two more: `OnLostFocus` never fires, so nothing may depend on being told focus *left*; and `OnSetFocus` reports "suppressed" while granting focus anyway, so it is not a usable predictor.

**Avalonia focus landing on a document view is a window-activation echo.** It names whichever document held focus last, not the one just clicked. Measured in a three-way split in a floating window: it named the same stale book every time, landing within ~15 ms of the correct CEF report, so a single click targeted the right book about half the time while a second click always worked. `IBrowserDocumentView` marks the views whose focus CEF speaks for, and the Avalonia feed ignores those landings — losing nothing, since that is exactly the set CEF reports directly.

## A correction to the record

`RootDock.FocusedDockable` is **not** "never populated in this app", as the resolver's comment claimed. It is populated on the **inner** root — Dock's `SetFocusedDockable` resolves via `FindRoot`, the nearest one — while every call site passes the **outer** `Root`. That is also why the repeated-⌘W nudge in `CloseDockableIfClosable` never worked: it writes where nobody reads. The activation feed repairs that as a side effect, since a close promotes a neighbour through the `ActiveDockable` setter.

## Scope: what this does not cover

With focus inside a book, CEF consumes the keystroke and the page relays it to its own view, so none of this is consulted — the target is then decided purely by which browser holds keyboard focus. That path has its own defect, **filed as #633**, and it is upstream of everything here.

## Testing

Full suite **1577 passed, 0 failed**, including 22 new tests: the tracker's recording rules, the resolver's new tier (including the reported sequence, and history walking past tools and other windows' documents), the echo rule, and a guard asserting that assigning `ActiveDockable` really does raise `ActiveDockableChanged` — if a Dock upgrade moves that call, the feed goes silent with nothing to show for it.

Verified in the running app on macOS: the PDF body no longer hijacks the keyboard, ⌘W closes what you are looking at, and Go To targets correctly across a three-way split in a floating window. **Windows unverified.**

Closes #621.

### DIFF
--- a/src/CST.Avalonia.Tests/Services/ActiveDocumentFeedTests.cs
+++ b/src/CST.Avalonia.Tests/Services/ActiveDocumentFeedTests.cs
@@ -1,0 +1,107 @@
+using CST.Avalonia.Services;
+using Dock.Model.Core;
+using Dock.Model.Mvvm.Controls;
+using Xunit;
+
+namespace CST.Avalonia.Tests.Services;
+
+/// <summary>
+/// Feed A of the interaction history (#621): the dock model's own activation signal.
+///
+/// <para>
+/// One subscription in <c>CstDockFactory</c>'s constructor is supposed to cover every dock in every window,
+/// including docks a split creates later and docks in floating windows. That claim rests entirely on a
+/// framework behaviour — Dock's <c>ActiveDockable</c> SETTER calling <c>InitActiveDockable</c>, which raises
+/// <c>ActiveDockableChanged</c> on the owning factory. If a Dock upgrade ever moves that call, the feed goes
+/// quiet with nothing to indicate it, and targeting silently reverts to the first-dock guess this fixes.
+/// These tests are the tripwire.
+/// </para>
+/// </summary>
+public class ActiveDocumentFeedTests
+{
+    private sealed class TestDocument : Document
+    {
+    }
+
+    private static (CstDockFactory Factory, ActiveDocumentTracker Tracker, DocumentDock Dock, TestDocument A, TestDocument B)
+        Build()
+    {
+        var factory = new CstDockFactory();
+        var tracker = new ActiveDocumentTracker();
+        factory.DocumentTracker = tracker;
+
+        var a = new TestDocument { Id = "a", Title = "A" };
+        var b = new TestDocument { Id = "b", Title = "B" };
+        var dock = new DocumentDock
+        {
+            Id = "MainDocumentDock",
+            Factory = factory,
+            VisibleDockables = new System.Collections.ObjectModel.ObservableCollection<IDockable> { a, b }
+        };
+
+        return (factory, tracker, dock, a, b);
+    }
+
+    [Fact]
+    public void Activating_a_tab_records_it()
+    {
+        var (_, tracker, dock, a, _) = Build();
+
+        dock.ActiveDockable = a;
+
+        Assert.Same(a, Assert.Single(tracker.Recent));
+    }
+
+    [Fact]
+    public void Successive_activations_build_the_history_in_order()
+    {
+        var (_, tracker, dock, a, b) = Build();
+
+        dock.ActiveDockable = a;
+        dock.ActiveDockable = b;
+
+        Assert.Equal(new IDockable[] { b, a }, tracker.Recent);
+    }
+
+    [Fact]
+    public void A_dock_created_after_the_factory_is_covered_too()
+    {
+        // The whole reason the subscription is on the FACTORY rather than per-dock: splits and floating
+        // windows create docks long after the constructor ran, and each would otherwise need wiring that
+        // someone has to remember.
+        var (factory, tracker, _, _, _) = Build();
+
+        var later = new TestDocument { Id = "later", Title = "Later" };
+        var newDock = new DocumentDock
+        {
+            Id = "MainDocumentDock",
+            Factory = factory,
+            VisibleDockables = new System.Collections.ObjectModel.ObservableCollection<IDockable> { later }
+        };
+
+        newDock.ActiveDockable = later;
+
+        Assert.Same(later, Assert.Single(tracker.Recent));
+    }
+
+    [Fact]
+    public void The_activation_event_carries_the_dockable_that_became_active()
+    {
+        // Pins the shape of the event args the feed reads, not just that something fired — reading the
+        // wrong member would record null and fail silently, which is the same as the feed not existing.
+        var factory = new CstDockFactory();
+        IDockable? seen = null;
+        factory.ActiveDockableChanged += (_, e) => seen = e.Dockable;
+
+        var doc = new TestDocument { Id = "x", Title = "X" };
+        var dock = new DocumentDock
+        {
+            Id = "MainDocumentDock",
+            Factory = factory,
+            VisibleDockables = new System.Collections.ObjectModel.ObservableCollection<IDockable> { doc }
+        };
+        dock.ActiveDockable = doc;
+
+        Assert.Same(doc, seen);
+    }
+}
--- a/src/CST.Avalonia.Tests/Services/ActiveDocumentTrackerTests.cs
+++ b/src/CST.Avalonia.Tests/Services/ActiveDocumentTrackerTests.cs
@@ -1,0 +1,149 @@
+using System;
+using System.Runtime.CompilerServices;
+using System.Linq;
+using CST.Avalonia.Services;
+using Dock.Model.Core;
+using Dock.Model.Mvvm.Controls;
+using Xunit;
+
+namespace CST.Avalonia.Tests.Services;
+
+/// <summary>
+/// The interaction history behind #621.
+///
+/// <para>
+/// It records order and nothing else — no filtering, no notion of what a document is. That is the design:
+/// the resolver filters on read by asking a layout what it contains, which is what lets one flat history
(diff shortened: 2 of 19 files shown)


fsnow:
Independent pass plus a Fable adversarial review. **Two findings fixed on the branch in `574f5b5`; one filed as #635.** Not merged yet — over to you.

## Verified rather than trusted

Based on current `main`. **1577/1577** exactly as claimed, and again after the fixes. No forbidden word in the diff.

Fable cleared the three axes that make this subsystem dangerous, and did real work to do it — it decompiled Dock 11.3.6.5 to check the framework behaviour Feed A rests on, and mutation-tested the suite (deleting the on-read containment filter fails four tests, so they bite):

- **Concurrency** — only Feed B arrives off the UI thread, and `CstWebView.OnGotFocus` marshals via `Dispatcher.UIThread.Post` before touching state. The tracker is lock-guarded and `Recent` returns a materialised snapshot.
- **Lifetime** — capacity 8 enforced on the only growth path, dead refs pruned on `Note`/`Forget`/`Recent`, collected-between-check-and-read handled. Weak entries root nothing.
- **CEF re-parent** — recreation goes through `FindControl` on the XAML-declared control, so the `is CstWebView` check survives float/unfloat and #458's invariant is untouched.

It also disproved the regression I was most worried about: a dead or failed browser is **not** invisible to both feeds, because its clicks land on child elements that aren't `IBrowserDocumentView`, so Feed C still records them.

## Fixed — `Forget` ran on a declined close

The history eviction sat between `base.CloseDockable` and the "did it actually leave" guard, with the comment explaining that guard four lines below it. A declined close dropped a still-open document from the history. Narrow today, since the reachable close paths pre-check `CanClose` — but it is precisely the guard-placement mistake **§E.1** exists to prevent, written a day after §E.1 was written. Both cleanups now share one `actuallyLeft` check.

## Fixed — `BrowserGotFocus` was never unsubscribed

The comment claimed the subscription "follows the browser's lifecycle for free" because `TryCreateWebView` re-runs after each dispose-and-recreate. But `FindControl` returns the **same** `CstWebView`, and `Dispose()` does not clear a C# event's handler list — so every window change stacked another lambda on the same delegate. Bounded (duplicate `Note`s are deduped; view and webview die together, so nothing is rooted across objects) but real. Named handlers now, detached in `DisposeWebView`, in `BookDisplayView` and `PdfDisplayView`. `WelcomeView` is deliberately untouched — it has no dispose path, so nothing can stack.

## Filed, not fixed — #635

**The echo rule suppresses by landing *type*, but the echo is defined by its *cause*.** `NoteFocus` walks to the first element with an `IDockable` DataContext and suppresses only if that element is an `IBrowserDocumentView`. A tab item is not one — and `SimpleTabbedWindow.cs:1184` already says *"Clicking a tab focuses the tab item, whose DataContext IS the dockable."*

So: click book A's tab, activate another window, click into book B's body. The restore records **A** unsuppressed while CEF reports **B**, and per this PR's own ~15 ms measurement the order is a coin flip.

I checked Fable's suggested repair and it is wrong: asking the *dockable* instead of the landing element would silence genuine tab clicks, which this PR explicitly depends on (no activation fires when the tab is already active in its own split). The real distinction is causal, and both viable directions touch focus around live CEF WebViews on macOS-only evidence — #633 territory. #635 records the sequence, why the obvious fix fails, and both directions.

Worth noting the PR's measurement wasn't wrong, just incomplete: it saw `source=BookDisplayView` because the browser had been the last focus in that experiment. The tab-item-last-focused configuration wasn't in the sample.

## One observation for #633

The split by document type is sharper than #633 currently states. `PdfDisplayView` has **zero** `CST_*` relay handlers against `BookDisplayView`'s **13**, so every PDF shortcut goes through layout resolution — which this PR now owns end to end. #633's residual risk is **book-to-book only**; a PDF can no longer be shadowed by a book. That matches the behaviour fsnow reported improving most.

Issue #623, opened 2026-08-13
Books that were floated or moved to another split are not reopened at startup


fsnow:
Reported by fsnow. Session restoration reopens books that stayed in the original DocumentDock, but silently drops any book that was floated into a separate window or dragged to a different split.

## Not a missing feature — the state is being deleted

Book state is registered per book at open time into `ApplicationState.BookWindows`, not gathered by walking the dock tree at save. So a floated book's entry already exists. It is then **deleted when the book moves**, because `CstDockFactory` calls `RemoveBookWindowState` from four places and only one of them is a genuine close:

| Site | What it is | Fires on a move? |
|---|---|---|
| `CstDockFactory.cs:143` | `CollectionChanged` on the main DocumentDock | **Yes** |
| `CstDockFactory.cs:1585` | `CloseDockable` override — the real close path | No |
| `CstDockFactory.cs:1818` | `RemoveDockable` override | **Yes** — Dock reuses it for reparenting |
| `CstDockFactory.cs:1944` | `CollectionChanged` on each floating window layout | **Yes** |

Dock reuses the same removal operation for "moved" and "closed", so none of the three event-driven sites can tell them apart. `SplitToWindow` calls `RemoveDockable` before re-adding to the new window; a cross-dock `MoveDockable` does the same.

Books that move *into* the main dock happen to survive, because setting `ActiveDockable` there triggers a re-save. Books landing anywhere else get no re-add.

## A second defect underneath it

`SaveAllBookWindowStatesAsync` walks `FindDocumentDock()`, which returns the **first** DocumentDock in the main tree — never a floating window, never a second split. So even once the deletes stop, a moved book's entry survives but never *updates*: reading position, hit index, `IsSelected` and `TabIndex` all freeze at the moment it left the original dock.

## And tab order is a stub

`BookWindowState.TabIndex` is written as a hardcoded `0` with `// TODO: Get actual tab index from dock` (`CstDockFactory.cs:695`) and never read on restore. Because `UpdateBookWindowState` is remove-then-add, `BookWindows` is in *last-touched* order, so books come back shuffled.

## Scope

Restore already opens every saved book into the main window's DocumentDock next to the Welcome page (`App.axaml.cs` `RestoreBookWindowsAsync`), which is the wanted behaviour. **Restoring the layout itself — which window, which split — is deliberately out of scope.**

## Follow-up considered and deferred

[...]

PR #637, opened 2026-08-14
Fix focus and dialog-owner targeting in splits (#633, #635)

### LINKED ISSUE #633: Clicking into a book does not reliably give its browser keyboard focus: window activation restores focus to another document view

### LINKED ISSUE #635: Window-activation echo can still be recorded when focus restores to a tab item
Found by adversarial review of #634 (PR for #621), verified in the code. **Narrows a fix rather than breaking one** — #634 is a large net improvement and this is the residue.

## The rule as built

`DocumentFocusReporter.NoteFocus` walks up from the focused element to the first one whose `DataContext` is an `IDockable`, then suppresses the report if **that element** implements `IBrowserDocumentView`:

```csharp
internal static bool ShouldReport(object? focusedElement) => focusedElement is not IBrowserDocumentView;
```

Only `BookDisplayView`, `PdfDisplayView` and `WelcomeView` implement it.

## Why that is the wrong key

The echo is defined by its **cause** — window activation restoring focus to whatever element held it last — not by the **type** of element it lands on. And the codebase already documents the case that slips through, at `SimpleTabbedWindow.cs:1184`:

> *"Clicking a tab focuses the tab item, **whose DataContext IS the dockable**."*

A tab item is not an `IBrowserDocumentView`, so the walk stops there and the landing is recorded.

## Failing sequence

1. In floating window W, click book **A**'s tab. Focus is now A's tab item.
2. Activate another window.
3. Click into book **B**'s body in W.

W activates, focus-restore raises `GotFocus` with the source being **A's tab item**, which is recorded as `avalonia-focus` → A. CEF reports B via `Dispatcher.Post`. Per #634's own measurement the two land within ~15 ms in **either order**, so the head of the history is a coin flip and ⌘W closes A about half the time.

This is the exact failure #634 was built to kill, re-entering through the one landing type the marker does not cover. The PR's measurement observed `source=BookDisplayView` only because the browser had been the last focus in that experiment; it does not cover the tab-item-last-focused configuration.

## The obvious fix is wrong

Suppressing by asking the **dockable** whether its view is an `IBrowserDocumentView` — rather than asking the landing element — would silence genuine tab clicks too. #634 lists those as a reason the Avalonia feed exists at all: no activation fires when the tab is already active in its own split, so a tab click is the only signal for that case.

The distinction needed is *user click on a tab* versus *activation restore landing on a tab*. That is causal.

## Directions, both needing measurement

- **Suppress at the cause:** ignore Avalonia-feed landings for a short window after `Window.Activated`, or set a flag across the activation-restore turn and clear it afterwards. Closest to the real distinction; introduces a timing dependency.
- **Suppress the restore itself** for views whose focus CEF owns — the option #633 calls "less invasive", and `IBrowserDocumentView` now supplies the scope for it.

**Do not do this casually.** Both touch focus around live CEF WebViews, which is the class of change behind several crashes and workarounds in this subsystem (`docs/architecture/DOCK_WEBVIEW_WORKAROUNDS.md`, #111 on sticky CEF focus). #634's measurement was macOS-only and Windows is now shipping.

## Related

Upstream of this is #633 — real keyboard focus landing on the wrong browser, which the in-page relay then faithfully honours. Both share the same underlying race (activation restore vs CEF's focus callback); this issue is about which of them the *history* believes, #633 is about which one the *keyboard* obeys.

Note the split by document type: PDFs have no in-page relay (`PdfDisplayView` has zero `CST_*` handlers against `BookDisplayView`'s 13), so every PDF shortcut goes through layout resolution and is fully governed by this path. Books get intercepted by the relay first.

## Testing note

`ShouldReport` is tested in isolation, but the `NoteFocus` walk that decides **which element the predicate sees** — the crux here — is not. A test driving the walk with a tab-item-shaped source would have caught this.

fsnow:
Four defects, all found by measuring rather than reasoning — three of them introduced by #621, and the one that actually caused the reported symptom older than any of them.

## The dialog owner — the reported bug

`FindWindowContainingBook` examined **one** document dock per window: the first. A split has several, so a book in the second pane came back "not found in any window" and its dialog opened owned by the **main** window.

That is not cosmetic. An owner in another window takes **activation** with it, and on macOS the click that brings activation back to the book's own window is consumed by activating it — never reaching the browser. So the next single click did not focus the book, and the shortcut after it acted on whatever browser still held CEF focus.

Reproduced with three books in a split: the first two worked, and the one whose dialog had just closed did not. The maintainer had independently noticed the dialog "first opened over the floating book, then later over the main window" — the visible half of the same bug.

It now asks whether **any** pane of a window holds the book, reusing the containment check resolution already performs. This is the same "first dock in tree order" assumption that #443 and #621 each fixed elsewhere; this was its third home.

## #635 — the echo rule was keyed on the wrong thing

Suppressing by the **type** of element focus landed on was wrong: the echo is defined by its **cause**, and an activation restore onto a tab item walked straight past the filter.

Avalonia carries the cause. `FocusManager`'s restore focuses with `NavigationMethod.Unspecified`; a click uses `Pointer`. So the rule reads that:

```csharp
internal static bool ShouldReport(NavigationMethod method) =>
    method is NavigationMethod.Pointer or NavigationMethod.Tab or NavigationMethod.Directional;
```

No timer, no flag across a turn, and no landing type can smuggle an echo through. `IBrowserDocumentView` had no consumer left and is deleted. The walk that decides *which* dockable the rule applies to — untested before, and exactly what #635 turned on — now has six tests, including the tab-item shape that defeated the old rule.

## #633 — the focus record

Avalonia cannot see a click on a browser's native surface, so its record stays on whatever was focused before, and the activation restore keeps resurrecting it: self-renewing staleness that the resolver's most trusted tier reads.

**Clearing was tried first and cannot work** — `ClearFocus()` is `Focus(null)`, which clears the current focus but not the per-scope element the restore reads, which lives in a private attached property only a successful focus overwrites. Observed directly: the same book was released at deactivation three times running while the user read two others.

So the view now focuses itself when CEF reports its browser took focus. **Measured safe rather than assumed**: in the same run Avalonia focused a `BookDisplayView` while the in-page relay went on delivering keystrokes from that browser, so view focus and CEF keyboard focus coexist — this records where focus is without taking it from anywhere.

## A crash on quit, from #621's browser-focus feed

Shutdown disposes the `ServiceProvider` while CEF still reports focus as browsers tear down. The queued callback called `ServiceProvider?.GetService`, which guards null but **not** disposal, and the `ObjectDisposedException` aborted the process.

`App.TryGetService<T>()` returns null instead, and every tracker lookup goes through it. Worth recording that the `Dispatcher.Post` added for thread-safety is what let the callback outlive the disposal — belt-and-braces bought a crash.

## Also: Go To resolves against the active window

On macOS the application menu claims the gesture, so a shortcut pressed while a floating window is frontmost arrives at the main window's handler, which answered from the **main** layout. With every book dragged out to a floating window that was the Welcome tab, and Go To did nothing at all.

## Diagnostics kept

Every interaction note records which feed produced it, and focus events log the book and the navigation method. Each round of this investigation turned on a fact the previous log could not distinguish — twice a plausible fix was built on a wrong guess and the log overturned it.

## Testing

Full suite **1604 passed, 0 failed**. Verified in the running app: dialog position and single-click focus are both correct across a three-way split in a floating window, and quitting no longer crashes.

**Windows unverified**, as with #621.

Closes #633. Closes #635.


fsnow:
Independent pass plus a Fable adversarial review. **One HIGH fixed on the branch in `21f754a`**; three doc-level items folded in. Merging.

## Verified

Based on current `main`. **1604/1604** as claimed, **1607** after the fixes. No forbidden word. `IBrowserDocumentView` is fully removed — no code or doc references survive. Nothing here touches `CloseDockable`, state deletion, or guard placement, so §E.1 and #623 are unaffected.

## Fixed — the #633 alignment was dead code on Welcome

`AlignFocusWithBrowser` focuses the view **root**, and Avalonia's `CanFocus` requires `Focusable`. `BookDisplayView.axaml:12` and `PdfDisplayView.axaml:11` set it on their roots; **`WelcomeView` set it on its inner web view instead**. So aligning the Welcome tab did nothing, and the PR's headline invariant failed on one of its three surfaces.

Not cosmetic — it is a wrong-tab action:

> click into book A → click into the Welcome page → press ⌘W → **book A closes**

because the resolver's most trusted tier is live Avalonia focus, still naming A since the alignment that should have moved it was a no-op.

Aligning to the inner web view would have been the wrong repair: that is the control the CEF adapter turns `GotFocus` into `SetFocus(true)` on, so it would push focus **into** the browser — the steal-back this design exists to avoid. `Focusable` now sits on the root.

**The failure was invisible because `Focus()`'s return value was discarded**, so the code read as though it maintained state it had silently stopped maintaining. It is checked and logged now, and a test asserts the property for all three views — mutation-verified: removing `Focusable` from the Welcome root fails it and nothing else.

## Also folded in

- **The `ShouldReport` doc was already false in-tree.** It claims programmatic focus is always `Unspecified`; `OpenBookPanel.axaml.cs:60` focuses a list item with `NavigationMethod.Directional`, which this rule reports as a user interaction. Harmless today — tools are filtered on read — but the next programmatic `Focus(Tab)` on a *document* would be a phantom interaction the rule exists to exclude.
- A detached view now returns early rather than falling through to a `Focus()` that cannot work, so "skip when already focused" does not silently invert to "always call" during teardown.
- `RecentDocuments`' summary was stacked on `CurrentLayout`.

## Verified clean, with better evidence than the PR claims for itself

Fable decompiled Avalonia 11.3.6 and CefGlue 120.6099.207 rather than reasoning from the description, and the risky change comes out **better founded than its own "measured on macOS" justification**: the only Avalonia→CEF focus edge is `AvaloniaControl.GotFocus → BrowserHost.SetFocus(true)` on the WebView control itself, there is **no** `LostFocus → SetFocus(false)` anywhere, and `GotFocus` bubbles *up* — so focusing an ancestor view never reaches the WebView control. No steal-back and no re-entrant loop, on any platform, not just the one it was measured on.

Also confirmed: the restore really does focus with `Unspecified` and the pointer handler with `Pointer`, so #635's rule holds on any landing element; `App.TryGetService<T>()` genuinely closes the disposal race rather than narrowing it (UI-thread serialization plus the `ObjectDisposedException` catch); and the dialog-owner fix covers every pane of every window, with reference-equality containment so duplicate instances of one book resolve to their own window.

## Left alone

`FindWindowContainingBook`'s window iteration is still untested — the tested `IsDocumentOf` helper is its load-bearing half. And **Windows remains unverified**, as with #621; given #518's precedent of a macOS measurement not transferring, the self-focus deserves one check on Placid or Kingfisher.

Issue #602, opened 2026-08-11
Surface B: a character-budgeted passage window can span ~30 paragraphs while the citation names one


fsnow:
Found by running the #583 orchestrator end to end against the real corpus and a real model (2026-08-11).

## What happened

A **Translate** turn on `s0502m.mul.xml` at **paragraph 21** (Dhp 21, the opening of the *appamādavagga*) produced a translation running from Dhp 21 through the end of chapter 2, the whole of **chapter 3** (*cittavagga*), and into **chapter 4** (*pupphavagga*) — roughly thirty verses.

Beside it, the app rendered:

> **tipiṭaka (mūla)/sutta piṭaka/khuddaka nikāya/dhammapadapāḷi, paragraph 21 (kn2)**

The model did nothing wrong: it translated exactly the text it was given, and the answer is good. The **citation is what misrepresents** — and the citation is the one thing on screen the app vouches for, rendered from its own data precisely so a garbled answer cannot forge one.

## Why

The passage window is budgeted in **characters** (`AiContextBundler.PassageBudget`, 2400 for Translate) and is structurally blind. It starts at the reference and flows on until the budget runs out. That is reasonable for prose — 2400 characters of a sutta is a passage. For **verse** it is not: a Dhammapada verse is ~80 characters, so 2400 characters is ~30 verses spanning multiple chapters.

The existing signal understates this badly. `BudgetReport.WindowMayExtendPastReference` was set and #582 surfaced its notice —

> The passage window may extend past the end of the cited reference.

— which is true, and reads like a rounding caveat rather than "this covers twenty-nine paragraphs you did not ask about". §6 names truthful output over a **narrower** scope than the user assumed as injection's characteristic failure; this is the same failure mirrored, and the citation makes it worse rather than better.

## Options

1. **Clamp the window to the cited reference** — take the paragraph and stop. Simplest and most honest; costs context on genuinely short paragraphs, which is what the surrounding-context argument for a character budget was for.
2. **Budget in paragraphs, not characters**, with a character ceiling as a backstop.
3. **Describe the window's real extent in the citation** — "paragraphs 21–49" rather than "paragraph 21". Needs the passage reader to report where it actually stopped, which is the unbuilt capability `ContextBundle.cs` already names.

(1) and (3) compose: clamp by default, and when the window does run on, say how far.

This also unblocks §6's fence 4 — the "partial passage" badge that #580 could not implement, because the same missing capability (did the window cover the reference?) is what drives it.

## Note for the eval set

[...]

Issue #546, opened 2026-08-01
View Source PDF opens the wrong volume for 12 books whose Myanmar page numbers restart


fsnow:
Found while implementing #540. Distinct from that fix, and user-visible today.

## The defect

`BookDisplayViewModel.ShowSource` parses the Myanmar page label and **discards the volume prefix**, keeping only the page number:

```csharp
var parts = _myanmarPage.Split('.');
if (parts.Length >= 2 && int.TryParse(parts[1], out int pageNum))
    currentPage = pageNum;          // "2.0001" -> 1, volume dropped
```

For 12 canonical books the Myanmar numbering **restarts at 1 in the second volume**, so every page in the second half collides with a page in the first. `s0201a` runs vol 1: 1–398, then vol 2: 1–398... er, 1–320 — and has a single source entry (`Mūlapaṇṇāsa-aṭṭhakathā -1.pdf`, PageStart 16). Reading in the second volume and pressing 1957 opens **page 16 of volume 1** — plausible-looking, entirely wrong, with no error.

## Affected books

| book | Myanmar volumes | source entries |
|---|---|---|
| `s0201a.att.xml` | 1: 1–398; 2: 1–320 | Burmese1957 @16 |
| `s0201t.tik.xml` | 1: 1–394; 2: 1–324 | Burmese1957 @16 |
| `s0502a.att.xml` | 1: 1–447; 2: 1–456 | Burmese1957 @8 |
| `s0505a.att.xml` | 1: 1–314; 2: 1–324 | Burmese1957 @4 |
| `s0510a.att.xml` | 1: 1–352; 2: 1–303 | Burmese1957 @4 |
| `s0513a4.att.xml` | 4: 1–504; 5: 1–113 | Burmese1957 @8 |
| `s0517a.att.xml` | 1: 1–345; 2: 1–323 | Burmese1957 @6 |
| `vin01a.att.xml` | 1: 1–345; 2: 1–312 | Burmese1957 @15 |
| `abh08t.nrf.xml` | 0: 1–138; 1: 139–360; 2: 1–378 | none |
| `vin07t.nrf.xml` | 1: 1–362; 2: 1–322 | none |
| `vin08t.nrf.xml` | 1: 1–424; 2: 1–434 | none |
| `vin11t.nrf.xml` | 1: 1–571; 2: 1–530 | none |

[...]

Issue #540, opened 2026-07-31
Source PDF: pages numbered in print but absent from the scan break the linear page mapping


fsnow:
Part of the #422 cluster. Raised as an explicit goal for View Source PDF: it must land on the right page **including where the print carries page numbers that the scan does not contain**.

## The assumption that breaks

`BookDisplayViewModel.ShowSource` maps a book page to a PDF page arithmetically:

```csharp
int pdfPage = source.PageStart + (currentPage - 1);
```

This assumes the scanned PDF contains **every** printed page, in order, with no gaps — one book page per PDF page throughout. Where a page is numbered in the printed edition but missing from the scan, every page after the gap is off by the number of missing pages, and the error grows with each subsequent gap.

It fails the same way as any other fixed-offset model: silently, and worse the further you read.

## Two possible solutions (per Frank)

1. **Skip programmatically** — record the missing pages per source PDF and adjust the mapping, so the arithmetic becomes "printed page → PDF page" through a gap table rather than a single offset.
2. **Insert blank placeholder pages into the PDFs** — restore the one-page-per-printed-page invariant in the data, so the existing linear mapping stays correct and no code changes.

Option 2 keeps the code simple and fixes it for every consumer at once, but means editing the preserved source PDFs. Option 1 leaves the PDFs untouched and puts the knowledge in `Sources.cs`, at the cost of a per-PDF gap map that has to be discovered and maintained.

Worth deciding deliberately — the choice determines whether `Sources.cs` grows a gap model or stays a single `PageStart`.

## Related work needed first
- Identifying **where** the gaps are. Unknown today; discovering them is its own task (and likely overlaps the #502 offset work, since both require inspecting the actual PDFs).
- Interacts with #76 (a book spanning several PDF volumes): a book that is both volume-split *and* gapped needs both models.

## Why it matters
View Source is a scholarly cross-check — the user is confirming the digital text against the printed page. Landing on the wrong page is worse than not offering the feature, because it silently invites a false verification.

fsnow:
## Confirmed in the wild: AN aṭṭhakathā, sub-book 2

First concrete instance, found while verifying #542 on macOS.

Jumping to **sub-book 2 in the AN 2-3-4 aṭṭhakathā**: the **Myanmar page number was exactly correct** — so anchor resolution (#542) is doing its job — but View Source landed **one page off**, because the `+18` offset no longer holds past a **missing blank page** in the scan.

[...]

Issue #446, opened 2026-07-21
Ranged paragraph @n produces WRONG paragraph citations in /v1/occurrences and /v1/passage


fsnow:
Found while scoping the #187 Phase 3 anchor catalog. This is the **live, shipped** consequence of the model limitation filed as #444 — filing separately because #444 is future design work, whereas this is wrong output today.

## The bug

`BookMarkers.Build` indexes a paragraph only when its `@n` parses as an integer (`src/CST.Core/Search/BookMarkers.cs:53`):

```csharp
case "p":
    if (!closing && attrs.TryGetValue("n", out var nStr) && int.TryParse(nStr, out int pnum))
        m._paras.Add((tag.Index, pnum, CurrentBook(divStack)));
```

Ranged values (`n="16-26"`, `n="196-7"`) fail `TryParse` and are silently skipped. `RefsAt` then answers with the last *indexed* paragraph at or before the position — which is the one **before** the range.

So a hit inside a ranged paragraph is attributed to the wrong paragraph. In `vin02t.tik.xml` the numbered sequence runs `13, 15, 16-26, 35, 36`: any hit inside `16-26` is reported as **paragraph 15**.

This is not a missing citation, it is an incorrect one — the caller has no signal that anything was approximated.

## Blast radius

`RefsAt` feeds both agent-facing read paths:
- `TeiSnippetExtractor.cs:118` → `refs` on `/v1/occurrences`
- `TeiPassageReader.cs:75,113` → `normalizedReference` on `/v1/passage`

Also `PositionOfParagraph(N)` returns -1 for any N inside a range, so paragraph addressing into those blocks fails outright.

## Scale (measured across the installed corpus)

**87 files, 3,618 paragraphs** carry a non-integer `@n`. Worst affected:

| file | ranged | total | share |
|---|---|---|---|
| abh08t.nrf.xml | 460 | 1028 | 44% |
| abh03t.tik.xml | 162 | 486 | 33% |
| abh05t.nrf.xml | 152 | 450 | 33% |
| abh03a.att.xml | 250 | 850 | 29% |
| vin06t.nrf.xml | 125 | 826 | 15% |
| s0505a.att.xml | 127 | 906 | 14% |
| vin11t.nrf.xml | 404 | 3165 | 12% |

Two range styles: full (`16-26`) and abbreviated (`196-7`, meaning 196-197).

## Suggested fix

Parse the range and index it, so a ranged paragraph is a first-class marker:
[...]

PR #638, opened 2026-08-15
Report every page a passage window covers (#561)

### LINKED ISSUE #561: /v1/occurrences and /v1/passage report different pages for the same hit at a page boundary
Found by a cold agent in the 2026-08-02 prompt-6 round (#260), then reproduced by hand.

## The defect

For the same cursor, the two endpoints disagree about which printed page the text is on — but only when the hit sits just after a page break.

```bash
POST /v1/occurrences {"bookId":"s0201a.att.xml","term":"kiñcāpi","take":4}
POST /v1/passage     {"bookId":"s0201a.att.xml","cursor":<each cursor>}
```

| cursor | `/v1/occurrences` refs | `/v1/passage` pages |
|---|---|---|
| 16053 | Myanmar 9 | Myanmar 9 |
| 99605 | Myanmar 54 | Myanmar 54 |
| 144980 | Myanmar 78 | Myanmar 78 |
| **145917** | **Myanmar 79** | **Myanmar 78** |

## Why it happens (hypothesis)

`occurrences` reports the page **at the hit**; `passage` returns a *window* and appears to report the page at the **window start**, which begins earlier and can therefore fall on the previous page. Both are defensible in isolation. Together they are not: the same text has two citations.

## Why it matters here more than in a typical API

- **Citation fidelity is the product.** #449 exists to steer agents to cite by page rather than paragraph, and `docs/testing/ai-prompts/citation-fidelity.md` tests exactly that. An agent that follows our own advice can now produce a citation that disagrees with what the reader sees.
- **It propagates into the source PDFs.** After #540, a printed page number maps to a PDF page. An off-by-one page reference lands the reader on the wrong scan — the precise failure that work removed.
- The disagreement is silent. Nothing in either response indicates the value is approximate or window-relative.

## Questions for the fix

1. Which is correct? A hit is a point, so `occurrences` reporting the page *containing the hit* looks right; `passage` reporting its window's first page is reasonable for a range but is not a citation for the hit.
2. If a window legitimately spans a page break, should `passage` return **all** pages it covers rather than one? It already returns an array, so the shape allows it — `pages` may simply be underpopulated at boundaries.
3. Whatever the answer, the two endpoints should agree for the same cursor, or the difference should be documented and visible.

## Reproduction

The commands above, verbatim, against a running instance with the local API enabled. Cursor 145917 in `s0201a.att.xml` is the failing case; the three others in the same result set agree, so any fix must keep those unchanged.

fsnow:
Closes #561.

## The disagreement

`/v1/occurrences` reports the page at the **hit**; `/v1/passage` reported the page at its **window start**, which begins earlier and can sit on the previous page. Both defensible alone — together they gave the same text two citations, with nothing in either response admitting it.

That matters more here than in a typical API. Citation fidelity is the product, `citation-fidelity.md` tests exactly this, and since #540 mapped printed pages onto PDF pages an off-by-one lands the reader on the wrong scan — the failure that work removed.

## Which endpoint was wrong? Neither, exactly

A hit is a **point**, so `occurrences` naming the page containing it is right. A window is a **range**, and it was describing itself by its first page only.

**#602 already settled the identical question for paragraphs, one line below.** `ReadWindow` calls `RefsAt(end - 1)` to report the paragraph in effect at the window's END, precisely so a caller can see the window's real extent — and it was discarding the pages from that very call with `_`. Pages now follow the same principle. `PassageResult.Pages` has always been a list; it was simply underpopulated.

`occurrences` is unchanged.

## What `PagesAcross` does

The page in effect at the start, then every break opening inside the span, per edition.

- **The first entry is unchanged**, so a caller reading `pages[0]` sees exactly what it saw before, and a window that crosses no break reports precisely what it did.
- A break exactly at the exclusive `end` belongs to the **next** window — otherwise consecutive windows both claim it and each overstates its extent.
- Editions stay grouped with each edition's pages in **reading order**, not sorted by number. Sorting numerically would misreport the twelve books of #546 whose page numbering restarts mid-volume.

## The scale was invisible from the reported cursor

Sweeping **12,508 cursors across all 217 books** through the real `TeiPassageReader`:

| | disagreements |
|---|---|
| before | **1,881** — roughly one passage in seven |
| after | **0** |

The four cursors named in the issue all agree, including **145917**, whose window now reports Myanmar 78, 79 and 80 with the hit's page 79 among them. The three that already agreed are unchanged, as the issue required.

## The invariant is a test, not a comment

> whatever `occurrences` cites for a hit must appear in what `passage` cites for a window covering it

Pinned at the unit level so a regression fails there first, and verified corpus-wide as above.

**1614/1614.**

### DIFF
--- a/src/CST.Avalonia.Tests/Search/BookMarkersTests.cs
+++ b/src/CST.Avalonia.Tests/Search/BookMarkersTests.cs
@@ -254,5 +254,25 @@ namespace CST.Avalonia.Tests.Search
                          pages.Select(p => p.Edition));
             Assert.Equal(new[] { 9, 3 }, pages.Where(p => p.Edition == PageEdition.Myanmar).Select(p => p.Number));
         }
+
+        [Fact]
+        public void An_edition_first_appearing_inside_the_span_can_take_the_first_slot()
+        {
+            // The case that makes "[0] is the page the passage starts on" FALSE, and the reason the doc
+            // comment now says so. Myanmar has a break before the window; Vri does not, so Vri contributes
+            // only the break inside it — and edition order puts Vri first. The entry at [0] therefore names
+            // a page the window's opening text is not on. Callers must read the list, or filter to the
+            // edition they mean. (fable review)
+            var xml = "aaaa<pb ed=\"M\" n=\"1.1\"/>bbbb<pb ed=\"V\" n=\"2.7\"/>cccc";
+            var markers = BookMarkers.Build(xml);
+            var start = xml.IndexOf("bbbb");
+
+            var atStart = markers.RefsAt(start).Pages;
+            var across = markers.PagesAcross(start, xml.Length);
+
+            Assert.Equal(PageEdition.Myanmar, Assert.Single(atStart).Edition);
+            Assert.Equal(PageEdition.Vri, across.First().Edition);      // NOT the same as RefsAt's first
+            Assert.Contains(across, p => p.Edition == PageEdition.Myanmar && p.Number == 1);
+        }
     }
 }
--- a/src/CST.Avalonia.Tests/Search/BookMarkersTests.cs
+++ b/src/CST.Avalonia.Tests/Search/BookMarkersTests.cs
@@ -1,3 +1,5 @@
+using System.Linq;
+using CST.Navigation;
 using CST.Search;
 using Xunit;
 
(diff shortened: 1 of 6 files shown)


fsnow:
Fable adversarial review. **Three findings fixed in `8fb9240`.** Two were mine and both were real.

## Fixed — the fix itself was untested

Every new test called `BookMarkers.PagesAcross` directly. **Nothing asserted that `TeiPassageReader` uses it.** Reverting the wiring left the whole suite green, because the one existing page assertion in `TeiPassageReaderTests` uses `Assert.Contains`, which a single-page result satisfies.

So the agreement between `/v1/occurrences` and `/v1/passage` — the entire point of this PR — could regress in a refactor with nothing to show for it.

Worth naming precisely what I got wrong: I *did* mutation-test this, but only through the scratchpad corpus sweep, which is not part of the suite. I checked the algorithm and assumed the wiring. Three tests now drive `ReadWindow` itself, and reverting the wiring fails two of them.

## Fixed — the compatibility claim was false

I wrote that "the first entry is the page in effect at the start, so a caller reading `pages[0]` sees no change." That holds only where the edition has a break **at or before** the window.

An edition whose first break falls **inside** the span contributes only that break, and edition grouping can then put it first:

```
aaaa<pb ed="M" n="1.1"/>bbbb<pb ed="V" n="2.7"/>cccc
```

A window from `bbbb` reports `[Vri 2.7, Myanmar 1.1]` where `RefsAt` reported `[Myanmar 1.1]`. `pages[0]` changes **edition and number**, and names a page the window's opening text is not on.

No in-repo consumer indexes `[0]` — checked `PassageTool`, `PromptBuilder`, `AnswerScorer`, `AiContextBundler` — so the exposure is external agents, plus a doc comment asserting something untrue. Both the comment and a new test now state the real contract.

## Fixed — the agent-facing contract said nothing about it

`llms.txt` and the MCP `passage` tool both described `pages` as though it were still one page per edition, while actively steering agents to **prefer citing a page**. Both now say it lists every page the window covers, that an edition can repeat, that within an edition the order is **reading order rather than numeric** — so a #546 restart book reporting Myanmar 9 then Myanmar 3 is correct rather than a bug — and that citing one exact point means using an `occurrences` hit's `refs`.

## Verified clean

Fable checked the boundary arithmetic exhaustively — `idx == -1`, a break exactly at `start` (counted once), a break exactly at `end` (excluded, so consecutive windows partition breaks with no double-report), empty edition lists, out-of-range spans, backwards spans — and found it correct, noting the `continue` can fire at most once. It also established the #561 invariant is **provable in general**, not merely true of the fixture: for any hit in `[readStart, end)`, its governing break either *is* the start-page break or lies strictly inside the span.

Ordering, performance (binary search then a scan bounded by the 20,000-char window cap), and the interaction with #446 all came back clean. Five of the seven original tests bite; the two weak ones are single-edition, which is exactly the gap the new multi-edition test fills.

**1618/1618.**

PR #640, opened 2026-08-15
Name an unknown request key instead of ignoring or refusing it namelessly (#558)

### LINKED ISSUE #558: /v1: a misnamed request key fails silently or namelessly — agents can't tell what they got wrong
Found in the 2026-08-02 cold-agent round (#260) — by an agent making the mistake naturally, which is the point of those runs.

`/v1` has **two different bad reactions to the same user error**, and neither names the offending key.

## 1. `navigate` ignores an unknown key silently

Sending `highlight` instead of `terms` — a plausible guess:

```bash
POST /v1/navigate  {"bookId":"s0301m.mul.xml","highlight":"mettā"}
→ HTTP 200  {"presented":true,"error":null,"bookId":"s0301m.mul.xml","highlights":0,"note":null}
```

The book opens, nothing is highlighted, and **nothing indicates the key was wrong**. Compare the benign case, which *does* explain itself:

```bash
POST /v1/navigate  {"bookId":"s0201m.mul.xml","terms":"mettā"}      # correct key, no matches
→ HTTP 200  {"presented":true,...,"highlights":0,
             "note":"No occurrences of 'mettā' in this book; no highlights were applied."}
```

So **the case that is the agent's own error gets no diagnostic, while the harmless case gets a clear one.** An agent reasonably concludes the phrase is absent from the book and moves on, having silently failed the user's "show me this passage" request. This is the only endpoint with a visible side effect, so a silent no-op is felt by the person at the screen.

## 2. `search` rejects a bad filter key with an empty body

```bash
POST /v1/search  {"query":"mettā","filter":{"nosuchkey":true}}
→ HTTP 400, body length 0
```

The 400 is deliberate and documented — `/docs/search.md` says a misnamed key is rejected *"so a misnamed key can't silently return the unfiltered corpus"*, which is exactly right. But the response says nothing, so the agent knows only that something in the request was unacceptable.

`/v1/occurrences` already does this properly:

```bash
POST /v1/occurrences  {"bookId":"nope.xml","term":"mettā"}
→ HTTP 404  {"error":"Unknown book 'nope.xml'."}
```

## Why it matters more for us than for a typical API

A human reads the docs, mistypes once, and sees the 400 in a console. An agent has no such loop — an uninformative failure is indistinguishable from a semantic result, and the friction reports show agents *reasoning onward* from the wrong conclusion rather than retrying. The whole surface is built so a cold agent can self-correct; these two paths remove that ability at the exact moment it is needed.

## Proposed fix

1. **`navigate`**: reject unknown body keys with a 400 naming them, or — if silent tolerance is wanted — emit a `note` when highlight-ish input was supplied under an unrecognised key. Preference for the 400: it matches `search`'s existing stance, and `note` is documented as "unexpected but non-fatal", which a wrong key is not.
2. **`search`**: give the existing 400 a body in the established shape, naming the key — `{"error":"Unknown filter key 'nosuchkey'. Valid keys: ..."}`.
3. Check the other `/v1` POST endpoints for the same asymmetry while we're in there.

Both are small. The contract already exists (`{"error": "..."}`); these two paths just don't use it.

## Evidence

All commands above were run against a live instance on 2026-08-02 and are reproducible as written. Round-1 cold-agent reports: `/tmp/cold-{opus,fable,sonnet,haiku}`.

fsnow:
Closes #558.

## The asymmetry pointed the wrong way

Found in a cold-agent round by an agent making the mistake naturally, which is what those runs are for.

```
navigate {"highlight":"mettā"}  → 200  highlights:0  note:null              ← the caller's error, silent
navigate {"terms":"mettā"}      → 200  highlights:0  note:"No occurrences…" ← harmless case, explained
```

**The agent's own mistake got the worse diagnostic of the two.** It concludes the phrase is absent and reasons onward — and `navigate` is the one endpoint with a visible side effect, so the person at the screen feels it. Separately, `search`'s deliberate 400 for a bad filter key carried a **zero-length body**.

## It was one default, not two bugs

There is no `UnmappedMemberHandling` anywhere in the repo, so `System.Text.Json` silently dropped unknown keys on **all eight** `/v1` POST endpoints: `search`, `occurrences`, `dictionary/lookup`, `passage`, `ai/context-preview`, `convert`, `navigate`, `forms`.

Patching the two reported paths would have left the other six — and the next endpoint added would inherit it. So the issue's own item 3 ("check the other endpoints for the same asymmetry") is answered by construction rather than by inspection.

Every POST body is now checked against its own contract before binding, and an unrecognised key gets a 400 in the established shape:

```json
{"error":"Unknown key 'highlight' in the request body. Valid keys: bookId, terms, …"}
```

Valid keys are read from the contract type by **reflection**, not hand-maintained, so they cannot drift from the endpoint.

## Two things measured rather than assumed

**`Disallow` alone gives the right status and an empty body.** My first attempt put the handler in exception-catching middleware — and the tests caught that it was dead code, because minimal APIs handle a binding failure internally instead of letting it reach the pipeline. .NET 10 has no `RouteHandlerOptions.ThrowOnBadRequest` to route it back out; **the type is not in the 10.0.9 reference assemblies at all.** So the check moved ahead of binding, where it is fully ours. `Disallow` stays as the backstop.

**Rejecting unknown keys is safe here in a way it would not be for a public API.** Loopback only, and an agent reads `llms.txt` from the *same instance* it then calls, so a client cannot be newer than its server. MCP is unaffected — those tools bind to the tool interfaces through DI, never over HTTP. And the entire existing suite passes unchanged, so nothing in-repo relied on extra keys being dropped.

## Testing

Against the **real server**, since the defect lived in the JSON/binding seam a mocked endpoint test steps straight over. Mutation-verified: bypassing the check fails five of the six.

One test is retargeted from `navigate` to `search` and says why — the fixture doesn't map `navigate`, which needs a presentation service. The defect was never navigate-specific, so driving it through a mapped endpoint tests the same thing.

`llms.txt` now states the rule where agents read it, including the actionable half: *if you get this, fix the key and retry rather than concluding the corpus lacks what you asked for.*

**1624/1624.**

### DIFF
--- a/src/CST.Avalonia.Tests/Integration/UnknownRequestKeyTests.cs
+++ b/src/CST.Avalonia.Tests/Integration/UnknownRequestKeyTests.cs
@@ -102,5 +102,78 @@ namespace CST.Avalonia.Tests.Integration
 
             Assert.Equal(HttpStatusCode.OK, resp.StatusCode);
         }
+
+        [Fact]
+        public async Task A_bad_key_INSIDE_the_filter_is_named_too()
+        {
+            // The issue's SECOND reported case, and the one a top-level-only check silently leaves broken:
+            // ToolBookFilter refuses unknown members itself, so binding already rejected this - with an
+            // empty 400, which is the exact diagnostic this change exists to replace.
+            var (status, body) = await Post("/v1/search",
+                "{\"query\":\"dhamma\",\"filter\":{\"nosuchkey\":true}}");
+
+            Assert.Equal(HttpStatusCode.BadRequest, status);
+
+            using var doc = JsonDocument.Parse(body);
+            var error = doc.RootElement.GetProperty("error").GetString()!;
+
+            Assert.Contains("filter.nosuchkey", error);
+            // The keys listed are the FILTER's, not the request's - naming top-level keys here would send
+            // the caller looking in the wrong place.
+            Assert.Contains("mula", error);
+            Assert.DoesNotContain("query", error);
+        }
+
+        [Fact]
+        public async Task A_valid_filter_still_works()
+        {
+            using var http = _api.Http();
+            var resp = await http.PostAsync("/v1/search",
+                Json("{\"query\":\"dhamma\",\"filter\":{\"mula\":true}}"));
+
+            Assert.Equal(HttpStatusCode.OK, resp.StatusCode);
+        }
+
+        [Theory]
+        [InlineData("/v1/convert", "{\"text\":\"dhamma\",\"outputScript\":\"Devanagari\",\"nosuchkey\":1}")]
+        [InlineData("/v1/dictionary/lookup", "{\"language\":\"en\",\"query\":\"dhamma\",\"nosuchkey\":1}")]
+        public async Task The_remaining_mapped_endpoints_name_the_key_too(string path, string body)
+        {
+            // The rejection is global (the Disallow backstop), but the NAMING depends on each entry in
+            // ContractFor. A typo'd entry regresses that endpoint to the bodiless 400 with every other test
+            // still green, so each mapped endpoint is asserted rather than assumed. (fable review)
+            var (status, text) = await Post(path, body);
+
+            Assert.Equal(HttpStatusCode.BadRequest, status);
+            Assert.Contains("nosuchkey", text);
+        }
+
+        [Fact]
+        public async Task An_unauthenticated_docs_path_does_not_reach_the_body_check()
+        {
+            // /docs is deliberately unauthenticated so a cold agent can orient itself. Matching the contract
+            // by path SUFFIX made POST /docs/search select the search contract, so an unauthenticated caller
+            // could make the server buffer an arbitrary body. Matching the full path closes it. (fable review)
+            using var http = new HttpClient { BaseAddress = new System.Uri(_api.BaseUrl) };   // NO token
+
+            var resp = await http.PostAsync("/docs/search", Json("{\"nosuchkey\":1}"));
+
+            Assert.NotEqual(HttpStatusCode.BadRequest, resp.StatusCode);   // not our unknown-key answer
+            Assert.Empty(await resp.Content.ReadAsStringAsync());
+        }
+
+        [Fact]
+        public async Task A_trailing_slash_or_odd_casing_still_gets_the_named_error()
+        {
+            // Route matching tolerates both; an Ordinal suffix match did not, so these fell through to the
+            // bodiless 400 this exists to remove. (fable review)
+            foreach (var path in new[] { "/v1/search/", "/v1/Search" })
+            {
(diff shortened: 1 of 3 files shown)


fsnow:
Fable adversarial review. **Four findings fixed in `0857f99`.** One of them means half the issue was still unfixed.

## Fixed — the issue's second case was still broken

#558's **case 2** is a *nested* key:

```
POST /v1/search {"query":"mettā","filter":{"nosuchkey":true}}  → 400, body length 0
```

My check read only top-level keys, so `ToolBookFilter`'s own `Disallow` rejected the body with **exactly the empty 400 this PR exists to replace**. I had written "give search's empty 400 a body" down as a task and marked it done without doing it.

Nested contract objects are now walked, reported by path — `filter.nosuchkey` — with the keys valid **at that level**. Listing the request's top-level keys for a bad filter sub-key would send the caller looking in the wrong place, so the containing contract is threaded through to the message.

## Fixed — suffix matching was wrong twice, once security-adjacent

`/docs` is deliberately unauthenticated so a cold agent can orient itself, and `EndsWith("/search")` matched **`POST /docs/search`** — letting an unauthenticated caller reach the body buffering. It also *missed* the real route whenever the path varied harmlessly (`/v1/search/`, `/v1/Search`), dropping those back to the bodiless 400.

Now matched on the full path, trailing slash trimmed, case-insensitively. Both the unauthenticated `/docs` path and the two variants are tested.

## Fixed — the middleware sat ahead of the concurrency cap

A request queued behind the 1024-deep limiter had already buffered its whole body — and that limiter exists precisely because this Kestrel shares a process with the Avalonia UI. It now runs after **both** the security gate and the limiter.

## Fixed — the naming was tested on three of eight endpoints

Rejection is global via the `Disallow` backstop, but the **naming** depends on each hand-typed entry in `ContractFor`. A typo'd entry regresses that endpoint to the bodiless 400 with every other test still green. The remaining mapped endpoints are now asserted directly.

Fable also caught that my excuse for retargeting the `navigate` test was false — `NavigateEndpointTests` wires navigate through the same harness with a `RecordingPresentationService`, so it was a convenience, not a constraint. Noted; the endpoint is covered by `ContractFor` and the exact-path tests.

## One failure was mine, and the feature reported it

I asserted a dictionary key of `word` when the contract says `query`. The new error named my mistake back to me — which is the whole point of the change.

## Verified clean

All eight `MapPost` routes are in the map, `/mcp` matches none of them and uses the SDK's own options so the global `Disallow` can't leak into MCP, and every contract is a plain positional record with no `[JsonPropertyName]` or inheritance — so the worst outcome, a **valid** request rejected because reflection disagreed with the binder, cannot occur. Rewind mechanics, chunked/empty/non-object/non-UTF8 bodies, and the existing integration suites all check out.

**1630/1630.**

Issue #691, opened 2026-08-18
UI: connections list and editor in AI settings


fsnow:
> ## How to read this issue
>
> **Match OpenCode.** The maintainer's direction, 2026-08-18: *"I had to fight with Kestrel today to get the UI implemented as I intended — as much like OpenCode as possible."* Screenshots and notes are in [`docs/reference/OpenCode/`](../tree/main/docs/reference/OpenCode). **Where this issue and OpenCode disagree, OpenCode wins unless the disagreement is marked [fsnow] below.**
>
> This issue is agent-authored, and its first version stated my inferences in the same confident register as the maintainer's decisions — which another session then implemented as rules, against his intent. So:
>
> - **[fsnow]** — his decision. Binding.
> - **[suggestion]** — mine. Advisory. Disagree freely; prefer OpenCode's answer if it differs.
> - **[observed]** — a fact read from the OpenCode study or from our code, with the source named.
>
> An unmarked sentence is context, not an instruction.

UI for #689. Binds to `IAiConnectionService` — see the seam in https://github.com/fsnow/cst/issues/689#issuecomment-5328504148. **Do not read `settings.json` or the keychain directly.**

Replaces today's single-provider block (one dropdown, one base URL, one key) with a list.

## The structure — two sections **[fsnow]**

Maintainer's direction, and the single thing most worth taking from OpenCode:

> Most importantly in OpenCode, I like the split between the configured providers at the top and the providers-to-add after that.

So the screen is **two stacked sections**, not one list (see `OpenCode Settings, Providers 2.png`):

1. **Your connections** — what is configured now. Empty on a fresh install, and that is fine; the section below is what the reader acts on first.
2. **Add a provider** — a catalogue of named presets, each with an **Add** action.

A provider that has been added **moves out of the lower section**, so the catalogue is always "what you could add next" rather than a list where some rows are already spoken for. Our current screen can only ever report on whichever provider the dropdown happens to be showing — this is the fix for that.

### What a preset is (and is not)

A preset supplies a **base URL** and a **key-required flag**, nothing else. It ships **no model list and no quality claim**, so it reintroduces nothing #670/#681 deleted — that was about *models*, not endpoints.

Its whole value is ergonomic: today a reader must already know `https://openrouter.ai/api/v1` to use OpenRouter at all. OpenCode's own list gives the trick away — its last row reads *"Custom provider: add an OpenAI-compatible provider by base URL"*, which is the generic case, and every named row above it is that same case with the URL filled in.

[...]

Issue #689, opened 2026-08-18
Epic: Surface B settings: configure providers and models separately, and support several endpoints


fsnow:
**Closed 2026-09-03 — delivered in Beta 6.** Every surface below shipped; see the closing comment for where the six remaining references live.

**Umbrella for the provider/model rework.** Filed because the design existed only as comments on #678 and #674 — which made it invisible to anyone scanning the issue list, and gave the work no place to be planned as a whole.

## The problem

Surface B's settings model is a single scalar provider: one `Provider` kind, one `BaseUrl`, one `Model` string, one credential. That was fine when the assumption was "configure a model once and read." It is wrong for how the feature is actually used.

The maintainer's read (2026-08-17), after seeing how cheap other tools make it to add an endpoint:

> Seeing how easy it is to configure new providers and models in OpenCode, I can see users working with multiple providers and models in CST Reader, and testing different models as they become available.

That is also exactly what the evaluation work needs — comparing two models on the same passage is the task, and it cannot be done while switching endpoints means editing one text box and overwriting one keychain entry.

## The shape

Three surfaces, at three different cardinalities — a dozen endpoints, hundreds of models, one choice per message:

1. **Connections** — a list, presets and custom endpoints alike, one row each. Add, edit, remove.
2. **Models** — per connection: which models are available, and which the reader wants on their short list.
3. **A per-turn picker** in the Assistant — the enabled models, grouped by connection, one click from the answer on screen.

### The connection record

```
Connections[] = { Id, DisplayName, Kind, BaseUrl, HasKey, Headers[], Models[] }
```

- **`Id` is a user-supplied slug**, immutable, and is what the keychain account is keyed on. Deliberately *not* derived from the base URL: a URL-derived key orphans the credential the moment someone changes a port or swaps `localhost` for `127.0.0.1`, and the resulting failure looks like a bad key rather than a lost one.
- **`Kind`** is ours — OpenAI-compatible or Anthropic. Note these are independent axes: several providers speak the Anthropic Messages protocol at custom URLs, so `Kind = Anthropic` must not imply `api.anthropic.com`.
- **`Models[]`** is a hand-entered list of `(id, display name)`. See the build order below.
- **The key is optional**, with `Headers[]` as the escape hatch — "leave empty if you manage auth via headers" covers Azure's `api-key`, gateway tokens, and anything non-bearer, and makes an absent key coherent rather than an error (which our resolver already believes for the local-runner case, but the UI never says).

### Build order — the part that changes the plan

[...]

Issue #673, opened 2026-08-17
Revisit the wait timeouts, and use published endpoint health instead of guessing


fsnow:
Companion to #672. Surface B was specced and built by an agent with little maintainer input, and the wait behaviour carries the same kind of unexamined numbers.

## What is currently assumed

`SseReader.DefaultIdleTimeout` = **120 seconds** — the gap allowed between lines once data is flowing.
`SseReader.DefaultFirstEventTimeout` = **10 minutes** — time to first data.
`HttpClient.Timeout` = infinite, deliberately, and guarded at construction; liveness is these two windows.
The selection round-trip through `document.title` = **700ms**.

The stated reasoning for the ten minutes is real and worth keeping in view: a local runner doing prompt evaluation on modest hardware can legitimately sit silent for minutes, and folding that into the idle window would abandon the fully-local path exactly when it is working hardest. That argument is sound *for a local runner*. It was then applied to every endpoint.

## Why it needs revisiting

**Ten minutes is not an interactive number.** A reader clicked a preset and watched nothing happen; the app was prepared to wait ten minutes before saying anything. The observed failure was a 120.5-second wait ending in a gateway 504, while a different request to the same model answered in 33 seconds. The panel now counts up and names queueing as the likely cause, but the ceiling itself is unchanged.

**One number serves two very different situations.** A loopback endpoint and a shared free tier have nothing in common. The local case is what earned the ten minutes; the hosted case is where a reader is left waiting. Keying the window to endpoint locality is the obvious split, and cheap: the base URL already distinguishes them.

**And for providers that publish health, waiting is not a guess at all.** Probing OpenRouter's `/api/v1/models/{canonical_slug}/endpoints` for the configured model returns four backends with `uptime_last_30m` of `null`, 100%, 99.0% and 97.5%, plus per-backend `max_completion_tokens` from 16,384 to 202,800. The caller does not choose the backend. So the app can say "this endpoint reported no uptime in the last half hour" while a reader is staring at a counter, instead of tuning a timeout to cover the worst case.

That is a better fix than any number: **report what is known rather than wait longer in case.**

## What to do

- Split the first-data window by endpoint locality — a short interactive ceiling for hosted endpoints, the existing long one for loopback and LAN.
- Make it configurable per provider rather than a constant, since the right value is a property of the endpoint and not of the app.
- Where the provider publishes endpoint health, surface it during the wait and after a failure. Vendor extension, not a standard, so enrichment only.
[...]

Issue #670, opened 2026-08-17
Remove the model registry and the fidelity advisory


fsnow:
Retire `ModelRegistry`, `model-registry.json`, the `FidelityAdvisory` binding, and the Settings card that renders it.

## Why

The registry maintains a per-model claim about fidelity on this corpus, surfaced in Settings as a tier. That is not a claim this project can keep true.

**Pāli ability is emergent from pre-training.** There is no Pāli-specific RL; what models can do here is capability overhang. So it is not predicted by published benchmarks, is not monotonic with general capability or size, and can move between point releases of the same model. There is no theory to interpolate from, which means a tier cannot be maintained by sampling — it would have to be re-measured for every release of every model. That is the "cannot", independent of the maintenance cost.

**The evidence cannot carry the word.** `docs/testing/ai-eval/cases.json` is 3 cases across 4 models. That supports "on these passages, on this date". "Recommended" implies a general property, and for translating a sacred text it is an endorsement the evidence does not earn.

**In practice it mostly renders a non-statement.** `Advisory` returns null for the recommended tier, so the card appears only for rated-but-not-recommended and unrated models. Almost every model a user types is unrated, so the common case is a panel announcing it has nothing to say.

**The one-model-per-family screening rule was already conceding this** — a sampling heuristic adopted because exhaustive measurement is impossible.

## Scope

Small. Nothing in the orchestrator, bundler, prompt or panel consults it; it never influences a request.

- `src/CST.Avalonia/Services/Ai/ModelRegistry.cs` (245 lines)
- `src/CST.Avalonia/Resources/Ai/model-registry.json` (77 lines, embedded resource)
- `src/CST.Avalonia.Tests/Services/Ai/ModelRegistryTests.cs` (177 lines)
- DI registration, `App.axaml.cs:1315`
- `FidelityAdvisory` / `HasFidelityAdvisory`, `SettingsViewModel.cs:1575`, raised at :1507
- the card, `SettingsWindow.axaml:612-618`
- the `IModelRegistry` parameter threaded into `AiSettingsViewModel`

## The doc must change too, or this regenerates

**`AI_INTEGRATION.md` §11.1 is the load-bearing part**, not the code. It instructs: *"Curate a recommended-for-translation frontier tier"* and *"a fidelity advisory when one is selected"*. `model-registry.json`'s `_comment` cites §11.1 as its warrant. Delete the code while the doc stands and a future session rebuilds it, correctly, on the doc's authority.

[...]

PR #698, opened 2026-08-18
The connection record and the provider preset catalogue (#689)

fsnow:
First backend increment of the provider/model rework (#689). **Deliberately additive** — nothing existing changes, `ChatSettings` keeps its scalar fields, the resolver is untouched.

Two reasons for that shape: it lets the UI work on kestrel (#691/#692/#693) bind against real types immediately rather than waiting on a migration, and it stayed clear of `AiCredentialStore` while the DPAPI work (#579, #690/#697) was landing.

## `AiConnection`

```
{ Id, DisplayName, Kind, BaseUrl, Models[], Headers, KeySource, State }
```

Three decisions worth reviewing:

**`Id` is user-supplied, immutable, and is the account the credential is filed under** — deliberately *not* derived from `BaseUrl`. A URL-derived key orphans the credential the moment someone changes a port or swaps `localhost` for `127.0.0.1`, and the failure then presents as a bad key rather than a lost one.

**`Kind` and `BaseUrl` are independent axes.** Several providers speak the Anthropic Messages protocol at their own hosts, so `Kind` must never be inferred from the URL. There's a test.

**`KeySource` and `State` are enums, not bools, because the third value is the point.** `CredentialSource.Environment` marks a credential we did not store and therefore cannot delete — so its row must offer no remove action, since an empty slot beats a button that lies. `Reachability.Configured` means *never contacted*, not *works*: conflating those is what lets a settings page say "Connected" while the assistant reports it cannot connect, which is precisely what OpenCode does and what #673 exists to avoid.

## `AiProviderPresets`

23 named endpoints, so a reader doesn't need to already know `https://openrouter.ai/api/v1` to use OpenRouter. A preset carries a base URL, a key-required flag, and the environment variables that conventionally hold a key — **no model list, no ranking, no quality claim** — so it reintroduces nothing from #670/#681, which concerned *models* rather than endpoints.

Facts come from `docs/research/PROVIDER_ENV_VARS_AND_ENDPOINTS.md`, mined from opencode (MIT) at `e14acea`, and **the commit is stamped in the file**. Refreshing means re-running the extraction, not hand-editing — the plan is to track upstream and fold in changes.

**Five providers are deliberately absent**, with reasons in the doc comment: Azure, Bedrock, Vertex, Cloudflare AI Gateway and Gemini each need fields the record doesn't model yet (resource name, SigV4, ADC, two tokens on one request, a third wire protocol). Guessing would ship endpoints that cannot work. All remain reachable by adding a custom endpoint by hand.

## Tests

15. The ones that earn their place are the **anti-curation guards**: no preset may carry ranking language in its display name, and ordering must be alphabetical rather than hand-arranged, because a reader reasonably reads "first" as "best".

Those matter most on a *future* sync, when whoever folds in upstream changes is reading a diff rather than the rules — models.dev carries pricing, `status` and vendor marketing copy, and that is how a curated tier would come back without anyone deciding to bring it.

Mutation-checked: making Ollama demand a key fails 1; renaming a preset to "Anthropic (recommended)" fails 1.

Suite: **1824 passed, 0 failed**, 0 warnings in both projects.

## Next

Not in this PR: `Connections[]` in settings, keychain re-keying by `Id`, `IAiConnectionService`, `CST_AI_*` overrides, and deleting the scalar `Provider`/`BaseUrl`/`Model` fields (safe to delete outright — `ChatSettings` postdates the beta 5 tag, so it has no users).

### DIFF
--- a/src/CST.Avalonia.Tests/Models/Ai/AiProviderPresetsTests.cs
+++ b/src/CST.Avalonia.Tests/Models/Ai/AiProviderPresetsTests.cs
@@ -1,0 +1,126 @@
+using System;
+using System.Linq;
+using CST.Avalonia.Models.Ai;
+using CST.Avalonia.Services.Ai;
+using Xunit;
+
+namespace CST.Avalonia.Tests.Models.Ai;
+
+/// <summary>
+/// #689: the preset catalogue is derived data — facts about third-party APIs, refreshed by re-running the
+/// #682 extraction against a newer opencode commit.
+///
+/// <para>These tests guard the two things that would quietly ruin it: a malformed entry (which ships an
+/// endpoint that cannot work), and <b>any drift toward curation</b>, which is the model registry removed in
+/// #670/#681 returning through a data file. The second matters most on future syncs, when someone folding in
+/// upstream changes will be looking at the diff rather than at the rules.</para>
+/// </summary>
+public class AiProviderPresetsTests
+{
+    [Fact]
+    public void Every_preset_is_well_formed()
+    {
+        Assert.NotEmpty(AiProviderPresets.All);
+
+        foreach (var p in AiProviderPresets.All)
+        {
+            Assert.False(string.IsNullOrWhiteSpace(p.Id), "a preset has no id");
+            Assert.False(string.IsNullOrWhiteSpace(p.DisplayName), $"{p.Id} has no display name");
+            Assert.True(Uri.TryCreate(p.BaseUrl, UriKind.Absolute, out _), $"{p.Id} has a non-absolute base URL");
+        }
+    }
+
+    /// <summary>Ids are the reserved namespace a custom connection may not take, and they become keychain
+    /// account names — so they must be unique and safe as a path/account segment.</summary>
+    [Fact]
+    public void Ids_are_unique_and_slug_shaped()
+    {
+        var ids = AiProviderPresets.All.Select(p => p.Id).ToList();
+        Assert.Equal(ids.Count, ids.Distinct(StringComparer.OrdinalIgnoreCase).Count());
+
+        foreach (var id in ids)
+            Assert.Matches("^[a-z0-9][a-z0-9_-]*$", id);
+    }
+
+    /// <summary>
+    /// A preset that requires a key but names no environment variable cannot participate in credential
+    /// discovery — the reader would be told nothing was found when a key was sitting right there.
+    /// </summary>
+    [Fact]
+    public void A_preset_that_needs_a_key_says_where_one_might_already_be()
+    {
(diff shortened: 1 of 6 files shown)


fsnow:
## Revised to follow opencode's own representation

Pushed `903de18`. Rather than inventing a shape, I read theirs in the local clone — since the plan is to track their provider work and fold in changes, a matching representation makes each sync a mapping rather than a re-interpretation.

### The change that matters

The first version had `RequiresKey: bool` + `EnvironmentVariables: string[]`. That turns out to be exactly opencode's `Integration.Method` `Key` and `Env` cases **flattened into two fields that cannot express a third** — and the third is the one we need. Vertex authenticates by ADC, Bedrock by SigV4 or the AWS credential chain; neither is a key.

`AiCredentialMethod` is now a union of `Key` / `Env` / `OAuth`, so those providers need credential machinery rather than a schema change. It also makes a rule structural instead of remembered: **`Env` says where a credential might be found; a connection exists only once the reader acts.**

### Also borrowed

- **`Inputs`** — the reader's own values (resource name, account id, region) that templates substitute. Azure and Cloudflare don't have a base URL so much as a shape.
- **`Prompts` with a `when` condition** — the preset declares which fields to ask for, and when. Azure wants a resource name *or* an explicit base URL; asking for both is wrong, and that rule now lives in data rather than hard-coded somewhere.

### Where we deliberately diverge

opencode carries the same URL templates but expands them in **per-provider plugin code** — `expandAccountId` at `packages/core/src/plugin/provider/cloudflare-workers-ai.ts:74` is a literal `replaceAll`, and every provider needing it writes its own. `AiTemplate` is one generic substitution: less code, and it cannot drift between providers.

Placeholders are `{key}`, matching the prompt that collects the value, rather than `${ENV_VAR}` — which conflates where a value came from with what it is called.

**An unresolved placeholder is left verbatim rather than emptied.** Emptying yields `https://.openai.azure.com/…`, which looks like a URL, survives a naive parse, and fails at request time as a DNS error naming nothing.

### Two of the five are now in

**Azure OpenAI** and **Cloudflare Workers AI**. Azure is what forced `AuthHeaderName`/`AuthScheme`: it sends the credential in `api-key` and expects `Authorization` to be **absent**, so adding a header wasn't enough.

Still out, with sharper reasons: **Cloudflare AI Gateway** (two independent credentials on one request), **Bedrock** (SigV4 signing), **Vertex** (ADC), **Gemini** (a third wire protocol — verified in the clone: opencode implements Gemini natively and has no OpenAI-compatible path for it).

23 tests now. Mutation-checked: emptying a missing placeholder fails 3. Suite **1838 passed**, 0 warnings.

Issue #693, opened 2026-08-18
UI: per-turn model picker in the Assistant


fsnow:
> ## How to read this issue
>
> **Match OpenCode.** The maintainer's direction, 2026-08-18: *"I had to fight with Kestrel today to get the UI implemented as I intended — as much like OpenCode as possible."* Screenshots and notes are in [`docs/reference/OpenCode/`](../tree/main/docs/reference/OpenCode). **Where this issue and OpenCode disagree, OpenCode wins unless the disagreement is marked [fsnow] below.**
>
> This issue is agent-authored, and its first version stated my inferences in the same confident register as the maintainer's decisions — which another session then implemented as rules, against his intent. So:
>
> - **[fsnow]** — his decision. Binding.
> - **[suggestion]** — mine. Advisory. Disagree freely; prefer OpenCode's answer if it differs.
> - **[observed]** — a fact read from the OpenCode study or from our code, with the source named.
>
> An unmarked sentence is context, not an instruction.

UI for #689. Binds to `IAiConnectionService` — seam in https://github.com/fsnow/cst/issues/689#issuecomment-5328504148.

The Assistant needs a way to change model **for the next message**, without a trip to Settings. This is the primitive the whole rework exists for: comparing two models on the same passage is the actual task, and it wants one click from the answer on screen.

## Shape

A **chip in the composer** showing the current model, opening a popup:

- a **search box** pinned at the top
- the **enabled** models (`AiModelEntry.Enabled`), **grouped by connection** under a group header
- the current selection marked (check or accent)
- **"Manage models…"** pinned at the bottom, routing to Settings

**[observed]** OpenCode's composer picker groups by provider under grey headers, with a pinned search box and `Manage models` at the bottom — see `OpenCode Chat, Provider-Model selector.png`. Match that. (Claude Desktop's is flat, which is why the first draft of this issue argued the point; OpenCode already settles it.)

## Behaviour

- The choice applies to the **next message**, and persists for the session.
- **[observed]** Selecting a model from a different connection switches credential and base URL with it — the service handles that (#678).
- **[suggestion]** Do not present a model as pickable when its connection has no usable key; show it disabled with the reason rather than letting the send fail with a 401 naming nothing. Advisory — OpenCode does not do this.
- **[suggestion]** Mark a connection whose `State` is `Unreachable`, but do not hide it — the reader may be about to start their local runner. Advisory; OpenCode shows no such state.

## Scope **[suggestion]**

[...]

Issue #692, opened 2026-08-18
UI: per-connection model list with enable toggles


fsnow:
> ## How to read this issue
>
> **Match OpenCode.** The maintainer's direction, 2026-08-18: *"I had to fight with Kestrel today to get the UI implemented as I intended — as much like OpenCode as possible."* Screenshots and notes are in [`docs/reference/OpenCode/`](../tree/main/docs/reference/OpenCode). **Where this issue and OpenCode disagree, OpenCode wins unless the disagreement is marked [fsnow] below.**
>
> This issue is agent-authored, and its first version stated my inferences in the same confident register as the maintainer's decisions — which another session then implemented as rules, against his intent. So:
>
> - **[fsnow]** — his decision. Binding.
> - **[suggestion]** — mine. Advisory. Disagree freely; prefer OpenCode's answer if it differs.
> - **[observed]** — a fact read from the OpenCode study or from our code, with the source named.
>
> An unmarked sentence is context, not an instruction.

UI for #689, after #691. Binds to `IAiConnectionService` — seam in https://github.com/fsnow/cst/issues/689#issuecomment-5328504148.

Each connection carries `Models: IReadOnlyList<AiModelEntry>` where `AiModelEntry = (Id, DisplayName, Enabled)`. This is the surface for turning entries on and off; the **per-turn picker (#692) shows only the enabled ones**.

## Shape

Grouped by connection, with a group header. Each row is the model's display name and a **toggle**.

**[observed]** OpenCode's control is multi-select: its composer picker shows several enabled models under one provider heading, which radio semantics could not produce. So `ToggleSwitch` or `CheckBox`, not `RadioButton` — a factual constraint, not a style preference. (The control was described both ways in discussion, hence the note.)

Put the **count in the group header**. It is exactly the number a reader needs to decide whether to expand, and it costs nothing.

## Default state

**[suggestion]** All-on when a connection is created.

The reasoning is worth having in front of you, because the obvious alternatives are both wrong in different ways:

- **all-on** — neutral. "Here is what this connection offers; prune it." **[observed]** this is what OpenCode does: its enabled list contains video and music models nobody would pick for chat, so a connected provider evidently arrives all-on and the user prunes.
- **all-off** — also neutral, but an empty picker with nothing to choose.
- **a pre-selected subset** — **[fsnow]** this is the one to avoid. Somebody chose which ones, which is the model registry deleted in #670/#681 wearing a toggle.

[...]

PR #706, opened 2026-08-18
Connections in settings, and the service the UI binds to (#689)

fsnow:
Stacked on #698 — review that first; this PR's diff is against it.

The half of #689 that unblocks the UI work on kestrel (#691, #692, #693) without waiting on credential plumbing.

## Settings

`ChatSettings` loses `Provider`/`BaseUrl`/`Model` and gains `Connections[]` + `ActiveConnectionId` + `ActiveModelId`. **Deleted outright rather than deprecated** — `ChatSettings` postdates the beta 5 tag, so there is no persisted state in the wild and no reader to break.

`AiConnectionRecord` is a separate *persisted* class from the `AiConnection` record in #698. The persisted one carries only what is stored; the runtime one adds what is *derived* (credential source, reachability), and neither belongs in a settings file.

## IAiConnectionService

What #691/#692/#693 bind to: connections, presets, available presets, active selection, add/update/remove, per-model enablement, and a `ConnectionsChanged` event.

**Deliberately no credential handling yet.** `KeySource` reports `None` and `State` reports `Configured` until the re-keying and reachability write-back land — both fields are already on the record, so neither addition changes a signature the UI touches. That is what makes the two halves parallel rather than sequential.

## Behaviour worth reviewing

- **A templated preset cannot be added without its inputs.** Azure without a resource name would keep its `{placeholder}` and fail later as a DNS error naming nothing; refused up front, naming the missing field.
- **Preset ids are reserved** against custom connections. The id becomes the credential's account name, so a collision would be one connection quietly inheriting another's key rather than a visible error.
- **Removing the active connection moves the pointer** rather than leaving it dangling — a stale id reads as "configured" to anything that only checks for null.
- **Models arrive enabled.** All-on is neutral; a pre-selected subset would be a quality verdict, i.e. #670/#681 wearing a toggle.

## The resolver

Now reads the active connection, and refuses one whose URL still has unfilled placeholders. A message naming the missing field beats a DNS failure naming nothing.

## The existing Settings UI still works

The single-provider fields were **not** deleted — they now edit the *active* connection. So the app stays configurable until #691 replaces them, and whatever a reader sets is a real connection record the new UI will show rather than state needing migration.

## Testing

20 new tests. **1858 passing**, 0 warnings in both projects.

## After this merges

All three UI issues are live, and the seam in #689 is real code rather than a proposal. Remaining backend: keychain re-keying by `Id`, `CST_AI_*` overrides, reachability write-back, and the three auth shapes.

### DIFF
--- a/src/CST.Avalonia.Tests/Services/Ai/AiConnectionServiceTests.cs
+++ b/src/CST.Avalonia.Tests/Services/Ai/AiConnectionServiceTests.cs
@@ -1,0 +1,253 @@
+using System.Collections.Generic;
+using System.Linq;
+using CST.Avalonia.Models;
+using CST.Avalonia.Models.Ai;
+using CST.Avalonia.Services;
+using CST.Avalonia.Services.Ai;
+using Moq;
+using Xunit;
+
+namespace CST.Avalonia.Tests.Services.Ai;
+
+/// <summary>
+/// #689: the service the connections UI (#691/#692/#693) binds to. Settings-backed; no credential handling
+/// yet, which is what lets the UI work start in parallel with the keychain re-keying.
+/// </summary>
+public class AiConnectionServiceTests
+{
+    private static (AiConnectionService Service, Settings Settings) Make()
+    {
+        var settings = new Settings();
+        var svc = new Mock<ISettingsService>();
+        svc.SetupGet(s => s.Settings).Returns(settings);
+        return (new AiConnectionService(svc.Object), settings);
+    }
+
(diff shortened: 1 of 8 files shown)


fsnow:
Superseded by #707 — same work, rebased onto `main`.

GitHub auto-closed this when #698's branch was deleted on merge; it was this PR's base, and a PR whose base branch disappears cannot be reopened or retargeted. My mistake for using `--delete-branch` on the parent of a stacked PR.

Issue #674, opened 2026-08-17
Model picker: choosing a model should not mean pasting a string


fsnow:
Switching models today means typing or pasting an id into a text box in Settings. For a project whose whole method is comparing models against Pāli, that is the wrong primitive.

The reference point is Claude Desktop's picker: a short list of the models you actually use, the current one checked, numeric shortcuts, and **More models >** for the rest. That works because there are four. It has to work where a provider lists **414** (measured against OpenRouter, 2026-08-16).

## The constraint that shapes it

**The short list must be user-derived, never curated by us.** A "recommended" list is precisely what #670 removes, and it would come back here wearing different clothes. Recents, and optionally pinned favourites — facts about what this reader has used, which need no maintenance and make no claim.

**The long list comes from the provider, not from us.** `ChatSettings.Model`'s comment says the id is *"never validated against a list and never a dropdown… any list shipped here would be wrong within a month."* That reasoning is against a **shipped** list. A **fetched** one is the opposite: self-updating, the provider's inventory rather than our assertion, and consistent with #670's distinction — asking the provider is fine, maintaining our own table is not.

**Free text must survive.** BYO-endpoint means a reader must always be able to type an id we have never heard of. The picker is an affordance, not a gate; a model absent from any listing must still be usable.

## What the provider gives us

`GET /api/v1/models` (OpenRouter) returns 414 entries with `name`, `context_length`, `pricing`, and `supported_parameters`. Enough to show a useful row rather than a bare id: name, context, whether it is free, whether it supports reasoning effort (#671).

Other providers differ. Anthropic exposes a models listing; Ollama and LM Studio expose `/v1/models` with far less metadata. So: render what is there, degrade to ids where that is all there is, and never require the listing to succeed before a reader can proceed.

## Open questions

**Where does it live?** The screenshot is a composer control, and switching models mid-session to compare answers on the same passage is a real use here — which argues for the Assistant panel, with provider, endpoint and key staying in Settings. Worth deciding deliberately.

**Search, not scroll.** At 414 a "More models" list is a search box, not a menu. Filter on id and name.

**Switching providers is not just switching models.** An id from a different provider needs that provider's base URL and key — see the per-provider credentials issue. A picker that silently switches model but not credentials would fail confusingly.

Related: #670 (no curated tiers), #671 (effort, discoverable from the same metadata).

fsnow:
## Prior art: OpenCode Desktop's Models page, and the shape it suggests

[...]

PR #725, opened 2026-08-19
Put a model's published facts on hover, not under every name (#674)

fsnow:
The id, context window, price and *reasoning* flag sat on a permanent second line under every model. At four
hundred rows that is four hundred lines of small grey text between the reader and the names they came to read.

They are now the row's **tooltip**. Nothing is lost and nothing is judged — the same provider-published facts,
verbatim, shown when the reader is choosing between two models. That is a moment, not a state, which is what a
tooltip is for.

The row becomes a name and a switch, with half the vertical padding it had, which also puts noticeably more of
a long listing on screen.

The **id gets its own line** in the tooltip rather than joining a run of dot-separated facts. It is the string
a reader would copy, and it is hard to pick out of `id · 131K context · free · reasoning`.

## What this costs

A tooltip is invisible until hovered, so **price is no longer readable at a glance while scanning** — and
price is exactly what the free/paid filter exists to act on. If that turns out to matter in use, the answer is
probably a small "free" marker on the row rather than restoring the whole line. Not doing that now: the
complaint was clutter, and adding something back before anyone has missed it is how clutter returns.

Search is unaffected — it still matches the id as well as the display name, so a model whose id is no longer
on screen is still findable by it.

## Rebased onto #723

Both touched this row template, and rebasing produced exactly the conflict predicted here. Resolved by keeping
the single `DisplayName` TextBlock and the `ToolTip.Tip` from this branch, and **main's two-column grid** —
the third column existed only for the "Typed" badge that #723 removed, so the toggle moves to column 1.

Deliberately not stacked on #723, since a follow-up pushed onto an open PR is what left the badge removal
orphaned when #722 merged without it.

## Testing

Clean build, 31 targeted tests on the Models tab, including that an endpoint publishing nothing yields a bare
id with no empty second line.

### DIFF
--- a/src/CST.Avalonia.Tests/ViewModels/AiModelsViewModelTests.cs
+++ b/src/CST.Avalonia.Tests/ViewModels/AiModelsViewModelTests.cs
@@ -518,7 +518,7 @@ public class AiModelsViewModelTests
     // ---- what a row says ---------------------------------------------------------------------------------
 
     /// <summary>The provider's own facts, verbatim and attributed — safe where a table we maintained would
-    /// not be.</summary>
+    /// not be. Carried on the row's tooltip rather than a second line of text under every name.</summary>
     [Fact]
     public void A_row_shows_the_published_facts()
     {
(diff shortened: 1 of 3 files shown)


fsnow:
Not merged — **this needs a rebase onto `main`.**

`AiModelsView.axaml` conflicts with #723 (the "Typed" badge removal), which landed after this branched. Both edit the model row's template, so git cannot reconcile them: #723 removed the badge from the row, this replaces the row's second line with a tooltip.

Not resolving it here — the correct result depends on what the row is meant to look like with both changes applied, which is a design question rather than a textual one, and you have the app in front of you.

Tested and merged tonight, for reference on what `main` now contains: #718 (subscription leak), #719 (startup crash), #720 (single scrolling region), #721 (enabling a model makes it active), #722 (free/paid filter), #723 (Typed badge removed), #724 (Manage models opens the right tab).

Rebase and I will batch it with whatever else is open.

PR #727, opened 2026-08-19
Match OpenCode's model card, and tell "no" apart from "no answer" (#693)

fsnow:
Side by side with OpenCode's card for the same Nemotron model:

| | theirs | ours, before |
|---|---|---|
| order | Model, Provider, Inputs, Reasoning, Context | Context and Reasoning reversed, id third |
| context | `1,000,000` | `1,000K` |
| reasoning | `Allows reasoning` | `supported` |

All three taken. The context window is a number readers compare exactly and the card has room for it;
"supported" does not read as a sentence on its own. The id row stays, moved last and still only when it
differs from the display name — it is the string a reader would copy, but not what they came to the card to
read.

## The distinction their card exposed

Theirs shows **"No reasoning"** for a local Ollama model. Ours could not have done better:
`SupportsReasoning` was a `bool`, so a provider that published a parameter list *without* reasoning in it and
a provider that published *no list at all* collapsed to the same `false`.

Those are different facts. A local runner's silence is not a statement about the model, and rendering it as
"No reasoning" asserts something nobody established — the same class of claim as reading an absent price as
free, which this code already refuses to make.

So it is `bool?` end to end: **null** says nothing, **false** says the provider said no, **true** says it said
yes. A test covers all three, and the Anthropic parser case now asserts null rather than false, because
Anthropic publishes no parameter list.

## Testing

**679 targeted AI tests pass, 0 warnings in both projects.** New: the three-state reasoning, and the card's
row order.

Worth checking in use — hover Nemotron on OpenRouter (should read exactly like their card) and a local Ollama
model (should show Model and Provider only, with no Reasoning row at all).

### DIFF
--- a/src/CST.Avalonia.Tests/Services/Ai/AiModelCatalogTests.cs
+++ b/src/CST.Avalonia.Tests/Services/Ai/AiModelCatalogTests.cs
@@ -52,7 +52,9 @@ public class AiModelCatalogTests
         var model = Assert.Single(models);
         Assert.Equal("Claude Sonnet 4.5", model.DisplayName);
         Assert.Null(model.ContextLength);
-        Assert.False(model.SupportsReasoning);
+        // Anthropic publishes no parameter list, which is not the same as publishing one without
+        // reasoning in it — so this is unknown rather than false.
+        Assert.Null(model.SupportsReasoning);
     }
 
     /// <summary>
--- a/src/CST.Avalonia.Tests/ViewModels/AiConnectionEditorViewModelTests.cs
+++ b/src/CST.Avalonia.Tests/ViewModels/AiConnectionEditorViewModelTests.cs
@@ -315,7 +315,7 @@ public class AiConnectionEditorViewModelTests
         var saved = Assert.Single(h.Service.Connections.Single().Models);
         Assert.Null(saved.ContextLength);
         Assert.Null(saved.Inputs);
-        Assert.False(saved.SupportsReasoning);
+        Assert.Null(saved.SupportsReasoning);
     }
 
     [Fact]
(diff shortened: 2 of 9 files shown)


fsnow:
Not merged — **needs a rebase**, and the cause is worth naming because it will keep happening.

This branch is stacked on **#726's pre-squash branch head** (`4b3a2f2`). #726 was squash-merged, so `main` has its content under a *different* commit (`5cd260e4`). Git therefore sees two unrelated histories that both changed the same files, and reports a conflict that is really just #726 appearing twice.

## The rebase that works

Replay only this branch's own commit onto `main`, dropping the one already merged:

```bash
git rebase --onto origin/main 4b3a2f2 feat/picker-card-fields
```

I ran that here. It leaves **one conflict, one hunk, in `AiModelsViewModel.cs`** — everything else applies cleanly.

I have not resolved it: the hunk is about how this PR's card fields combine with what #726 already landed, which is a design decision rather than a textual merge, and you have the app in front of you.

## For future stacked work

After a squash-merge, a branch stacked on the pre-squash head always needs `--onto`; a plain `git rebase origin/main` will try to replay the parent's commit again and conflict with itself.

Also worth knowing, since I hit it twice tonight: merging a parent PR with `--delete-branch` **closes** any PR stacked on it, and a closed PR whose base branch is gone cannot be reopened or retargeted. I have stopped passing `--delete-branch` for that reason.

## What `main` contains now

#718, #719, #720, #721, #722, #723, #724, #725, #726 — all tested and merged tonight. Current head `5cd260e4`, 2062 tests passing, app verified starting.

Rebase and I will batch it with anything else open.

Issue #739, opened 2026-08-19
UI: present ~166 providers without a flat list, and name a catalogue failure


fsnow:
UI (kestrel). Depends on #737 landing — before that, the provider list is 25 entries and none of this is needed.

> **Provenance.** Agent-authored. **[fsnow]** = binding. **[suggestion]** = advisory. **[measured]** = verified. Where this and OpenCode disagree, **OpenCode wins** unless marked **[fsnow]** — screenshots and notes in `docs/reference/OpenCode/`.

## What changes

**[measured]** The "add a provider" section goes from **25** entries to **~166**. A flat list stops working somewhere well below that.

**[observed]** OpenCode's answer, from `OpenCode Settings, Providers 1.png`: a short **Popular providers** section, then **"Show more providers"**. Their catalogue is the same 189-odd; that is how they present it.

## Wanted

- Progressive disclosure over the full set, plus **search** that filters across it.
- **[fsnow]** No ranking. So whatever fills a "popular" section cannot be a quality judgment. Honest options: the ones the reader has used before, alphabetical-with-search and no popular section at all, or a short set chosen on a stated mechanical basis. Note OpenCode's own popular section includes their first-party products — that is placement, not a signal we can borrow.
- **[suggestion]** Recently-used is a fact about this reader and ranks nothing, which makes it the safest basis if a short section is wanted.

## The failure state

**[suggestion]** #736 can fail to reach models.dev — offline, a proxy, an outage. The section must then say **"couldn't reach the provider list"**, offer retry, and keep the **custom endpoint** route visible.

An empty section reads as a broken feature. A stale list fails softly; a missing one fails hard, which is why this needs a named state rather than nothing.

Note the baked-in snapshot (#736) means this should be rare — but "rare" is exactly when a screen with no explanation costs the most.

## Local providers stay first-class

**[fsnow]** Ollama, LM Studio and the custom-endpoint route are baked in and work with no network at all. They must remain reachable when the catalogue is unavailable — arguably the case that matters most, since a local runner needs no internet to be useful.

## Not this issue

Logos and the documentation link — see the companion UI issue. The catalogue itself — #736, #737.

Issue #738, opened 2026-08-19
Provider logos from models.dev, with the monogram as fallback


fsnow:
Backend, with a small UI consumer. Part of #733; unblocks the logo half of #691.

> **Provenance.** **[fsnow]** = binding. **[suggestion]** = advisory. **[measured]** = verified.

## Why now

**[fsnow]** decided provider logos are in (#691). They were deferred to coloured monogram tiles because there was no asset source, and I had noted a licensing question about shipping third-party marks.

**[measured]** models.dev publishes them: `https://models.dev/logos/{provider}.svg`, keyed by the same provider id we already use, **MIT**, with a documented default served when a provider has none.

That answers both objections at once — a licensed set, keyed to data we already pull.

## What

A resolver that hands the UI a logo for a provider id, and a cache so it is fetched once.

**[suggestion]**

- Fetch on demand per provider, not all 192 up front. A reader sees a handful.
- Cache to disk beside the catalogue (#736); treat a miss as non-fatal.
- **The monogram stays as the fallback** — for a provider with no logo, for a custom endpoint (which has no provider id at all), and for any fetch that fails. It is already built and works; this adds a better first choice, it does not replace the mechanism.
- Local runners: **[suggestion]** keep the monogram. Ollama and LM Studio are not models.dev providers, and inventing marks for them is not our business.

## Consumers

The connections list and the add-a-provider catalogue (#691), and the model picker's group headers if it reads well there — **[suggestion]** that last one is Kestrel's judgment with the app in front of them, not mine from a diff.

## Deliverable

- Logo resolution + disk cache, offline-tolerant.
- Monogram fallback preserved on every path.
- The MIT notice carried for the asset set.
- Tests: cache hit, missing logo falls back to monogram, fetch failure does not surface as an error.

## Not this issue

Rendering decisions — size, placement, whether a logo appears in the picker as well as the list — belong to the UI issues.

fsnow:
Backend is built. One thing the UI issues need to know before they can consume it, which I did not decide.

## The resolver hands back a file path to an SVG, and the app cannot draw an SVG

**[measured]** Nothing in the tree renders SVG. There is no `Avalonia.Svg.Skia` or `Svg.Skia` package reference in any csproj, and the one place that mentions SVG — `OpenBookPanel.axaml:12`, `<!-- Mac-style SVG icons -->` — is a comment above an empty `<UserControl.Resources>`. Avalonia's `Image` takes a `Bitmap`; it has no SVG decoder.

[...]

Issue #737, opened 2026-08-19
Generate the provider presets from the cached catalogue


fsnow:
Backend. Depends on #736. Implements #733 §2–3, and **subsumes #731**.

> **Provenance.** Agent-authored. **[fsnow]** = binding. **[suggestion]** = advisory. **[measured]** = command given.

## What

Replace the hand-written `AiProviderPresets` table with presets derived from the cached catalogue.

**[measured]** Today it ships **25** entries, hand-picked by me from the ~70 the #682 spike surfaced. The maintainer's question — *"I don't know how we selected the providers that we do support"* — had no answer in the code, because the rule was never written down.

## Two independent axes

**[measured]** `api.json` holds 192 providers.

**Axis A — where the base URL comes from:**

| | count | source |
|---|---|---|
| has `api` | **166** | straight from the document |
| no `api` | **26** | hand-kept table; models.dev omits the URL when a dedicated `@ai-sdk/*` package carries its own default |

The 26 are **not** unsupported — eight are providers we ship today that work: `anthropic`, `openai`, `groq`, `cerebras`, `deepinfra`, `togetherai`, `xai`, `azure`. Filtering on `api` would drop OpenAI and Anthropic.

**This is #731's real cause.** The spike recorded "SDK default" because that is what the upstream data says, not through carelessness. Its fix — resolve each URL once from vendor documentation — becomes the hand-kept table here.

**Axis B — whether our resolver can authenticate it.** Independent of A, **[measured]** demonstrably:

| provider | has `api`? | needs |
|---|---|---|
| `snowflake-cortex` | yes | non-key credential |
| `databricks` | yes | host + token pair |
| `github-copilot` | yes | OAuth-ish token |
| `amazon-bedrock` | no | SigV4 (#702) |
| `google-vertex` | no | ADC (#703) |
| `cloudflare-ai-gateway` | no | two credentials (#701) |
| `azure` | no | **already supported** — template + `api-key` header |

## Emit rule

A preset is emitted when **A yields a URL** (either source) **and B is a shape we serve** — bearer, Anthropic's `x-api-key`, or Azure's `api-key` (#712).

**[suggestion]** Everything skipped is **logged with its id and reason**. My hand-picked list shrank silently from ~70 to 25 and nobody could tell why; a generated one must not repeat that.

## Constraints

[...]

Issue #529, opened 2026-07-28
AI settings: apply changes live instead of requiring a restart


fsnow:
Split out of #527, which covered only telling the user a restart is needed. That hint has shipped (#528); this issue is the real fix.

## Current behaviour

The AI settings are read **once, at startup**:

```csharp
/// changes to the setting take effect on restart (live toggle is a later iteration). (#186)
private async Task StartLocalApiIfEnabledAsync(ISettingsService settingsService)
{
    var ai = settingsService.Settings.Ai;
    if (!ai.ServerShouldRun) return;
    ...
    await _localApiServer.StartAsync();
```

`StartAsync` is called only from that path, `StopAsync` only at shutdown, and nothing subscribes to settings changes. So toggling **Enable AI Features**, **Run the local API server** or **Enable MCP access** does nothing until relaunch.

## Wanted

Toggling any of those starts or stops the server there and then — no restart, and no "⚠ Takes effect after you restart CST Reader." note needed in Settings.

## Why it is more than an event handler

The server owns three pieces of state that all have to move together:

1. **The Kestrel host.** Start and stop must be safe under repeated rapid toggling (a user flipping the checkbox), which means serialising transitions rather than racing `StartAsync`/`StopAsync`.
2. **The handshake file** (`local-api.json`). Written on start, deleted on clean shutdown. A live stop must delete it, and a live start must write a **new port and token** — both are per-session today, so anything that cached them is stale.
3. **The two surfaces** (`/v1` REST and `/mcp`). They are enabled independently, so the transitions are not just on/off: REST-only → both → MCP-only all have to work without dropping the surface that stayed on.

The hard case is **a client connected mid-flight**. Disabling AI while an MCP chat client or a coding agent is mid-request should fail that request cleanly rather than hang or crash, and the `--mcp-bridge` relay (#278) — which reads `local-api.json` each spawn — needs to behave sensibly when the file disappears under it.

## Suggested approach

- Subscribe to the AI settings section and funnel every change through one serialised `ApplyAsync(desiredState)` that diffs current vs desired and performs the minimum transition.
- Keep `StartLocalApiIfEnabledAsync` as the startup call into that same path, so there is one code path rather than two.
- On stop: delete the handshake file first, so a client that polls it stops finding a live endpoint before the port actually closes.
- Cover with tests for the transitions, including both-off → both-on and each single-surface flip.

## Related
- #527 (the UI hint, done)
- #186 (the epic where the "later iteration" note originates)
[...]

PR #741, opened 2026-08-19
Fetch and cache the models.dev provider catalogue (#736)

### LINKED ISSUE #736: Fetch and cache the models.dev catalogue
Backend. First of the #733 implementation issues — everything else there depends on this one.

> **Provenance.** Agent-authored. **[fsnow]** = maintainer's decision, binding. **[suggestion]** = mine, advisory. **[measured]** = read or fetched, command given. The design this implements is the first comment on #733.

## What

A service that keeps a local copy of `https://models.dev/api.json` and hands it to callers. Nothing in this issue changes what a reader sees — it is the supply.

## Shape

**[fsnow]** Store the document **whole**. Subsetting on ingest is not a safety control; the constraint lives on display and ordering (see #733 §8).

**[measured]** opencode's own mechanism, `packages/core/src/models-dev.ts`, is worth copying rather than reinventing:

- fetch `api.json`, write the **raw text** to the cache directory, temp-file then rename;
- load order **disk cache → baked-in snapshot → network**;
- **5-minute TTL** on file mtime as a freshness guard, background refresh every **60 minutes**;
- a **cross-process lock**, so two instances cannot race the cache file;
- a fetch failure **logs and keeps the previous copy**; a flag disables fetching entirely.

**[suggestion]** Model the C# on `DpdUpdateService` — version-stamped, hard timeout over the whole check — but much lighter: one JSON document, no releases API, no staged install. `api.json` is **[measured]** 4,008,548 bytes; use a conditional GET so an unchanged document costs a 304.

## When it runs

**[fsnow]** Gated on AI and the assistant being enabled. A reader who never turns AI on makes no network call for it.

**[suggestion]** Driven from the same live-apply path as #529, not startup alone — enabling AI mid-session should fetch immediately rather than at next launch.

## The build-time snapshot

**[suggestion]** Bake a copy of `api.json` into the app as the final fallback, as opencode does (`packages/opencode/script/build.ts:195`). Two reasons: a fresh install with no network still has providers, and it removes the empty-list failure state entirely.

Worth deciding whether the snapshot is refreshed by hand at release time or by a build step. Either is fine; **stamp what it was taken from**, as `AiProviderPresets.SourceCommit` already does.

## Deliverable

- `IModelsDevCatalog` (or similar): `Task<CatalogDocument?> GetAsync(CancellationToken)`, plus a refresh entry point.
- Cache under the app data directory, alongside the other derived assets.
- Failure is a **state the caller can render**, not an exception and not silence — see the UI issue for what that has to support.
- Tests: cache hit, TTL expiry, fetch failure keeps the previous copy, snapshot used when no cache exists, and a malformed document rejected without taking the app down.

## Not this issue

Turning the document into presets is the next one. Logos are separate again. Nothing here is user-visible.

Licence: models.dev is **MIT**; carry the notice with any snapshot shipped in the app.

fsnow:
Closes #736. First of the #733 implementation issues; #737 and #738 depend on it.

**Nothing user-visible.** This is the supply that #737 turns into presets.

## Two commits

**1. A committed snapshot, refreshed by its own script.** `refresh-models-snapshot.sh` fetches `api.json`, validates it on a temp file, and writes it pretty-printed with sorted keys.

Deliberately *not* in the package scripts — the maintainer named the deciding objection: `package-macos.sh` and `package-windows.ps1` are independent and each runs its own `dotnet publish`, so a fetch in one would force an ordering between them (macOS first, or Windows ships a stale copy) with nothing saying so. It would also make builds non-reproducible and turn a models.dev outage into a release failure.

Pretty-printed because a 4 MB minified diff is unreadable, and seeing which providers appeared or disappeared between releases is the whole reason for committing it. **5.7 MB expanded, ~0.4 MB as git stores it.** A sidecar `.meta.json` records source, fetch time, counts, sha256 and licence. `RELEASE_PROCESS.md` gains it as a pre-release item.

**2. The catalogue service.** Load order **cache → snapshot → network**, matching opencode's `models-dev.ts`.

## The property this rests on: a failure never destroys what we had

Every failure path keeps the previous copy and reports a `Problem` alongside it, because the alternative a reader sees is an empty provider list — which reads as a broken feature rather than a network problem (#739).

- unreachable → keep the copy, say so
- **a 200 carrying an HTML error page** → same. This is the realistic failure, not a 404, so the *shape* is validated rather than the status code trusted
- a corrupt cache → discarded **and deleted**; it would otherwise fail identically on every subsequent start
- no cache and no snapshot → a build fault, said plainly rather than presented as "no providers exist"

## When it runs

Gated on AI **and** the assistant being enabled — a reader who never turns AI on makes no network call for it.

Driven from `ApplyAiSettingsAsync`, **not startup alone**. Gating on "AI enabled" is right, but checking only at launch reproduces #529 exactly, where a mid-session toggle did nothing until relaunch. Fire-and-forget, since a checkbox must not wait on a network call. One-hour freshness window, so several launches in an hour cost one request.

## Tests

12. The ones that earn their place: a fresh install with no network still gets providers from the snapshot; **a provider without an `api` URL is still carried**, because reading its absence as "unsupported" would drop OpenAI and Anthropic (#737); and a corrupt cache is deleted rather than left to fail again — which caught a real gap, since `Parse` returns null instead of throwing so my `catch` never fired.

**2109 passing**, 0 warnings in both projects, app verified starting.

## Note

#733 §9 was corrected today on Kestrel's measurement: the catalogue does **not** replace a live `/v1/models` fetch. It trails OpenRouter by 64 usable models and does not list `ollama` at all. That concerns *models*; presets describe *endpoints*, where the catalogue is authoritative. Unaffected by this PR.


fsnow:
## Fable pre-merge review — seven findings, all fixed

Two would have failed in ordinary use.

**The poisoning path was the serious one**, and it is the single case where a failure destroyed what we had — the property this design rests on. A 200 carrying `{"error":{"id":"rate_limited"}}` is valid JSON, deserializes to one record with an id, passed `Count > 0`, **overwrote the good 192-provider cache**, and became the answer for every subsequent start; a reader who then went offline was stuck with one provider. My own refresh script demanded ≥50 providers of the same document while the runtime demanded ≥1. Now they match, and a test asserts both the in-memory copy and the file on disk survive an error body.

**The catalogue never refreshed at launch.** `ApplyAiSettingsAsync` has exactly one caller — the settings save — while startup reaches the lifecycle directly. So a reader who enabled AI once and never re-saved rode the same cache forever, and the freshness window was a dead letter. Worse, my comment claimed the *inverse* defect. Fixed and **verified by launching**: `Provider catalogue refreshed: 192 providers` now appears in a real startup log.

**A client timeout threw out of `RefreshAsync`** — `is not OperationCanceledException` excludes `TaskCanceledException`, which is what `HttpClient.Timeout` raises, so a hung endpoint would have thrown past a contract promising a `Problem`, straight through #739's retry button.

Plus: atomic script write with the temp in the trap, an honest `GetAsync` doc, gating on `Ai.Enabled` alone (it feeds the *providers* list, browsed before the assistant is enabled), and both failure tests asserting `Source == Cache` and the count rather than merely `NotEmpty` — as written they would have passed if the snapshot had silently replaced the cache.

## Verified clean by the review

The one I most wanted checked: **the embedded resource name resolves at runtime**, confirmed in the built assembly, and the snapshot tests genuinely exercise it — empty temp cache dir, a 2-provider stub network, so a count over 100 can only come from the embedded resource. Green for the right reason.

Also clean: the `_current` fast path (atomic reference read of an immutable record), every other failure path, script determinism, no ranking anywhere, and MIT attribution carried.

## Deliberately not changed

A malformed field nulls the whole parse rather than dropping one provider. The snapshot test runs the real parser over the real embedded bytes, so an unparseable snapshot fails the suite **before it ships** — per-record tolerance would convert that loud pre-release failure into a silent provider drop.

**2111 passing**, 0 warnings, app verified starting.

Issue #742, opened 2026-08-19
A bare-host base URL always gains /v1, which is wrong for at least one real provider


fsnow:
`AiHttp.ResolveEndpoint` adds a version segment to any base URL with no path:

```csharp
var addVersion = segments.Length switch
{
    0 => true,                                        // bare host always needs the version segment
    _ when IsVersionSegment(segments[^1]) => false,
    _ => convention == BaseUrlConvention.ExcludesVersion || segments.Length == 1,
};
```

That is right for most providers — `https://api.deepseek.com` really does serve `/v1/chat/completions` — and
it is why the rule was written. It now has a **measured counterexample**.

## The counterexample

**[measured]** from the vendor's own SDK, fetched without installing anything:

```bash
npm pack @ai-sdk/perplexity --pack-destination .
tar xzf ai-sdk-perplexity-*.tgz
grep -ohE 'baseURL\}[^`"]{0,30}' package/dist/index.js | sort -u
#   baseURL}/chat/completions
#   baseURL}/v1/embeddings
```

with `baseURL` defaulting to `https://api.perplexity.ai`. So Perplexity's chat endpoint is
**`https://api.perplexity.ai/chat/completions`** — a bare host that does *not* take a version segment. Note
its own embeddings route does, which is presumably how the convention came to differ.

Against that, `ResolveEndpoint` builds `…/v1/chat/completions`, and every request 404s.

## Why filing it separately

#737 excludes Perplexity from the generated presets, correctly — inventing a per-preset endpoint convention
while implementing the middle issue of a chain is scope creep, and it is recorded there as a logged skip.

**But that closes only the preset door.** A reader can type `https://api.perplexity.ai` into the custom
endpoint form today. It is a valid, documented base URL, pasted from the provider's own docs — and it produces
a connection that looks correct, whose model listing 404s and whose every question 404s, with nothing on
screen suggesting the URL was altered. The reader cannot see the `/v1` we added.

This is reachable now, without any of #733's work.

## Shape

Not proposing an implementation; the design belongs with whoever picks it up. The options as I see them:

- **A convention on the connection**, defaulting to today's behaviour and set by a preset where the provider
  needs it. Fits the existing `BaseUrlConvention` enum, which already exists for exactly this axis and is
  currently chosen from `Kind` alone.
- **Try, then fall back** — a 404 on `…/v1/chat/completions` retried once against `…/chat/completions`. Cheap
[...]

Issue #740, opened 2026-08-19
UI: show provider logos and a per-provider documentation link


fsnow:
UI (kestrel). Depends on #738 (logo assets) and #737 (which carries `doc` through to the preset).

> **Provenance.** **[fsnow]** = binding. **[suggestion]** = advisory. **[measured]** = verified. Where this and OpenCode disagree, **OpenCode wins** unless marked **[fsnow]** — see `docs/reference/OpenCode/`.

## 1. Logos

**[fsnow]** Provider logos are in. The monogram tiles were a stand-in for want of an asset source; #738 supplies one (models.dev SVGs, MIT, keyed by the provider id we already use).

**[suggestion]** Replace the monogram with the logo where one resolves, and **keep the monogram wherever one does not** — a provider with no logo, a custom endpoint (no provider id at all), a failed fetch, and the local runners. It already works; this is a better first choice, not a replacement mechanism.

## 2. The documentation link

**[measured]** Every one of the 192 catalogue records carries a `doc` URL. Sampling ten, nine point at a **models** page:

```
openrouter   https://openrouter.ai/models
openai       https://platform.openai.com/docs/models
anthropic    https://docs.anthropic.com/en/docs/about-claude/models
```

So it is **model documentation, not account or key documentation** — worth stating because my first note on this called it help for "a reader who does not know where their key comes from", which is wrong about who is on that screen. Anyone pasting a key has already been to the provider and generated one.

Its use is the reverse: a reader with a working connection who needs to know **what to run on it** — which model ids exist, at what price and context length.

**[suggestion]** Two placements follow from that:

- on a connection row or in its editor — *"see this provider's models"*;
- on the **Models tab** when a listing is empty, fails, or the endpoint publishes none. That last case is the one with no other answer today: a reader must type a model id obtained from somewhere, and this is the somewhere.

## Not this issue

Fetching or caching either asset — #738 for logos, #736/#737 for the catalogue. Presenting 166 providers — #739.

fsnow:
## Backend note from #737: the picker resolves presets statically

Raised here rather than fixed in #744, since it is UI-side.

`AiModelPickerViewModel.cs:310` uses `AiProviderPresets.ById(connection.Id)`, which after #737 reads the **build-time snapshot** rather than the live preset list held by `IAiConnectionService`.

[...]

PR #744, opened 2026-08-19
Generate the provider presets from the catalogue (#737)

### LINKED ISSUE #731: Presets omit mainstream providers because the spike recorded no base URL for them
Part of #689 / #691.

> **Provenance.** Agent-authored. **[fsnow]** marks the maintainer's words; **[suggestion]** is mine, advisory; **[observed]** is measured from the repo, with the command given.

## How this surfaced

**[fsnow]** *"I just discovered that OpenCode has a much larger list of supported providers that I had not seen. Actually I don't know how we selected the providers that we do support."*

Neither did anyone else, because the rule was never written down. I chose them, in #698.

## What the selection actually was

**[observed]** The #682 spike catalogued **~70 providers** into `docs/research/PROVIDER_ENV_VARS_AND_ENDPOINTS.md`. `AiProviderPresets` ships **25**.

The filter I applied, reconstructed from the code rather than from memory: include anything the research doc gave a **concrete base URL** for, that authenticates with a plain key, in one of our two wire shapes.

Four omissions were deliberate and are documented in the file, with issues: Bedrock (#702), Vertex (#703), Cloudflare AI Gateway (#701), Gemini (#700).

**About 45 were omitted silently**, and a significant group of those for a reason that is not a judgment at all: the research doc records their base URL as **"SDK default"**, meaning my extraction never captured a concrete URL. Reproduce with:

```bash
grep -c "SDK default" docs/research/PROVIDER_ENV_VARS_AND_ENDPOINTS.md
```

That group includes **Mistral, Cohere, Perplexity, Vercel AI Gateway** and **GitHub Copilot** — mainstream providers whose absence reflects a gap in my extraction, not a decision about them.

## Why this is a bug rather than a scoping choice

These endpoints are ordinary: a documented base URL, a bearer token, the OpenAI-compatible shape we already serve. Each is one row in the presets table. They are missing because of how the data was gathered, and nothing in the code says so — the file documents the four deliberate omissions and is silent about the forty-five.

## Suggested fix

**[suggestion]**

- Re-run the #682 extraction for the providers recorded as "SDK default", resolving each concrete base URL — from the vendor's own documentation where opencode delegates to an SDK.
- Add those that are a plain base URL plus a key. Anything needing extra credentials or a third wire shape gets an issue instead, as the first four did.
- **Record the inclusion rule in `AiProviderPresets`.** Its absence is why this was undiscoverable. Whatever the rule ends up being, a reader of that file should be able to tell why a provider is missing without asking.
- Re-stamp `SourceCommit`.

Note this is separate from the larger question of whether the list should be **hand-picked at all** — see the companion issue. If that one is taken, this may be subsumed by it; if it is not, this stands on its own.

### LINKED ISSUE #737: Generate the provider presets from the cached catalogue

fsnow:
Closes #737. Closes #731. Depends on #736 (merged).

Replaces a hand-picked table of **25** with **~170** derived from the models.dev document, plus a small hand-kept table for what the catalogue cannot supply.

The maintainer's question — *"I don't know how we selected the providers that we do support"* — had no answer in the code, because I picked them and never wrote the rule down. **The rule is now in the file:** a preset is emitted when we have a base URL, from the catalogue or the hand-kept table, **and** its credential shape is one the resolver serves.

## Why the hand-kept table is not a leftover

models.dev records an `api` URL exactly when a provider is served by the generic OpenAI-compatible adapter, and omits it when a dedicated SDK package carries its own default — so **26 of 192 have none, including OpenAI and Anthropic**. Emitting only what has an `api` field would drop both.

**The wire format is a column there, not an assumption.** Each entry carries an explicit `ChatProviderKind`, and that enum has two members — so a provider speaking a third protocol cannot be added even by accident. Cohere posts to `/chat` on `api.cohere.com/v2` and would need a value that does not exist. Kestrel measured that from the SDK package; **I would have recorded a base URL for it and shipped a preset that 404s on every request.**

## Four skip filters, every one from a real record

Found by running the generator over the real document rather than by imagining failure modes:

| skipped | why |
|---|---|
| `wafer.ai` | id is not a slug — and the id becomes the credential's keychain account name |
| 26 providers | no base URL |
| `databricks`, `snowflake-cortex`, `neon`, `infomaniak` | URL templated against env vars the catalogue expects the *host* to expand (`${DATABRICKS_HOST}`) — our templating fills from reader-supplied Inputs declared by **prompts**, which a catalogue record has none of |
| — | URL not an absolute http(s) address |

**Every skip is logged with its id.** Verified in a real launch:

```
Provider presets: skipped 17 with no base URL (add to AiProviderPresets.HandKept
to support one): aihubmix, amazon-bedrock, ... perplexity, ... watsonx (#737)
```

The old hand-picked list lost ~45 providers silently and nobody could say why.

`perplexity` is skipped for a different reason and is now **#742**: its base URL has no version segment, and `ResolveEndpoint` adds one to a bare host, so it would post to `/v1/chat/completions` against an endpoint serving `/chat/completions`. Not papered over with a doctored URL.

## State, not a bare list

Requested by Kestrel from the UI side **before this was built**, and correctly. Two of three outcomes arrive as an empty list: "all added" is a quiet end state, "couldn't reach the catalogue" needs a sentence and a retry, "not fetched yet" wants "loading" — and the loud one reads as a broken feature.

`AiPresetState { Loading, Ready, Unavailable }` + `PresetProblem` + `RefreshPresetsAsync`, surfaced through `IAiConnectionService`.

Two mappings worth reviewing: **the snapshot is `Ready`**, not a failure — a fresh offline install genuinely has those providers. And **`Unavailable` still returns the local runners**, since they need no network to be useful and are exactly the wrong thing to hide when the network is down. Both pinned by tests.

## Regressions I caused and caught

Rebuilding the table dropped **Cloudflare Workers AI** (it needs an `accountId` prompt), and `ById`/`IsReservedId` were still consulting only the hand-kept part — so `openrouter` read as unknown, which would let a custom connection claim the id and lose its key-required flag. Both found by the existing suite.

**2124 passing**, 0 warnings, app verified starting.

### DIFF
--- a/src/CST.Avalonia.Tests/Models/Ai/AiProviderPresetsTests.cs
+++ b/src/CST.Avalonia.Tests/Models/Ai/AiProviderPresetsTests.cs
@@ -61,7 +61,9 @@ public class AiProviderPresetsTests
     [Fact]
     public void A_preset_that_needs_a_key_says_where_one_might_already_be()
     {
-        foreach (var p in AiProviderPresets.All.Where(p => p.RequiresKey))
+        // Asserted of the HAND-KEPT entries only. Build treats the catalogue's `env` as optional, so a
+        // future snapshot carrying an env-less provider would break this suite rather than the app.
+        foreach (var p in AiProviderPresets.HandKept.Where(p => p.RequiresKey))
             Assert.NotEmpty(p.EnvironmentVariables);
     }
 
--- a/src/CST.Avalonia.Tests/Models/Ai/AiProviderPresetsTests.cs
+++ b/src/CST.Avalonia.Tests/Models/Ai/AiProviderPresetsTests.cs
@@ -70,7 +70,10 @@ public class AiProviderPresetsTests
     [Fact]
     public void Local_runners_do_not_require_a_key()
     {
-        var local = AiProviderPresets.All.Where(p => p.BaseUrl.Contains("localhost")).ToList();
+        // Keyed off OUR local presets rather than "any URL containing localhost": the catalogue lists at
+        // least one hosted provider (privatemode-ai) that advertises a loopback address, and it is not a
+        // local runner in the sense this rule is about.
+        var local = AiProviderPresets.LocalOnly;
 
         Assert.NotEmpty(local);
         Assert.All(local, p => Assert.False(p.RequiresKey, $"{p.Id} is local but demands a key"));
@@ -146,7 +149,7 @@ public class AiProviderPresetsTests
     [Fact]
     public void Local_runners_declare_no_credential_method()
     {
-        foreach (var p in AiProviderPresets.All.Where(p => p.BaseUrl.Contains("localhost")))
+        foreach (var p in AiProviderPresets.LocalOnly)
         {
             Assert.Empty(p.Methods);
             Assert.False(p.RequiresKey);
--- a/src/CST.Avalonia.Tests/Services/Ai/AiPresetSourceTests.cs
+++ b/src/CST.Avalonia.Tests/Services/Ai/AiPresetSourceTests.cs
@@ -232,4 +232,86 @@ public class AiPresetSourceTests
         Assert.Equal(1, catalog.Refreshes);
         Assert.Equal(1, raised);
     }
+
+    // ---- the wire protocol comes from the catalogue, not from an assumption (fable review) -------------
+
+    /// <summary>
+    /// The finding that would have shipped eight broken presets. Eight catalogue providers carry a perfectly
+    /// valid `api` URL and declare `@ai-sdk/anthropic` — minimax, thinkingmachines, kimi-for-coding and
+    /// others — so that URL serves Anthropic Messages, not `/chat/completions`. Probed live:
+    /// `/anthropic/v1/chat/completions` returns 404 while `/anthropic/v1/messages` returns 401.
+    /// </summary>
+    [Fact]
+    public void A_provider_declaring_the_anthropic_sdk_is_emitted_as_anthropic()
+    {
+        var built = AiPresetSource.Build(Catalogue(
+            new CatalogProvider("minimax", "MiniMax", "https://api.minimax.io/anthropic/v1",
+                Npm: "@ai-sdk/anthropic")));
+
+        var minimax = Assert.Single(built.Where(p => p.Id == "minimax"));
+        Assert.Equal(ChatProviderKind.Anthropic, minimax.Kind);
+    }
+
+    /// <summary>Deliberately not a general package-to-kind mapping: OpenRouter ships its own provider package
+    /// and is genuinely OpenAI-compatible, so only the one package whose protocol we can name is special.</summary>
+    [Fact]
+    public void Another_vendors_own_sdk_package_does_not_change_the_protocol()
+    {
+        var built = AiPresetSource.Build(Catalogue(
+            new CatalogProvider("openrouter", "OpenRouter", "https://openrouter.ai/api/v1",
+                Npm: "@openrouter/ai-sdk-provider")));
+
+        Assert.Equal(ChatProviderKind.OpenAiCompatible,
+            built.Single(p => p.Id == "openrouter").Kind);
+    }
+
+    /// <summary>Endpoints whose URL parses fine but which our resolution addresses wrongly. Both were
+    /// confirmed by probing rather than reasoned about; revisit when #742 lands.</summary>
+    [Theory]
+    [InlineData("github-copilot", "https://api.githubcopilot.com")]
+    [InlineData("perplexity-agent", "https://api.perplexity.ai/v1")]
+    public void An_endpoint_we_would_address_wrongly_is_not_emitted(string id, string api)
+    {
+        var built = AiPresetSource.Build(Catalogue(new CatalogProvider(id, id, api)));
+
+        Assert.DoesNotContain(built, p => p.Id == id);
+    }
+
+    /// <summary>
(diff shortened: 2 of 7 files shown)


fsnow:
## Fable review — nine findings, all fixed

Two would have shipped presets that could only fail, and the reviewer probed the endpoints rather than reasoning about them.

**The wire protocol is in the catalogue and I ignored it.** `Build` hardcoded `OpenAiCompatible` for every generated preset, but the `npm` field names the protocol: **eight** providers carry a valid `api` URL and declare `@ai-sdk/anthropic` — minimax, thinkingmachines, kimi-for-coding and siblings — so that URL serves Anthropic Messages. Probed: `/anthropic/v1/chat/completions` → **404**, `/anthropic/v1/messages` → 401.

The galling part is that the capability was already there: `AiProviderPresets` documents *"Kind must never imply this host"* and our Anthropic adapter resolves those URLs correctly. The generator simply never consulted the field.

**Two more emitted presets 404**, live-confirmed: `github-copilot` (bare host, serves `/chat/completions`, and can't authenticate with a pasted key anyway) and `perplexity-agent` — the sibling of the `perplexity` case this PR already skipped by hand, which walked straight through the filter.

**The collapse guard was dead code.** `hosted` subtracted only the local runners, but `Build` always seeds the whole hand-kept table, so it was permanently above zero. The reshape it was written to catch would have reported *Ready* with a dozen presets and no problem.

**The injected path was worse than no service at all** — the constructor seeded two local runners while `EnsureLoadedAsync` ran after the network refresh, so with a stale cache and a hung host, `AddFromPreset("openai")` answered *"not a known provider"* for an HTTP timeout's length.

**And the error handler broke this feature's one rule**, resetting `Presets` and discarding a good list.

## Second pass

I reported the documentation fixes as done when they were one-third done — my string replacements silently no-opped and I didn't assert. Redone with assertions, and two more taken:

**A tie-break on id**, which is the anti-registry rule rather than tidiness: `OrderBy` is stable, so a shared display name fell back to insertion order — the *source document's* order, and the last channel by which models.dev's ordering could decide what a reader sees first. No such pair exists today, which is precisely the wrong reason to leave it.

**An honest sentence for the second failure**: reached-and-unusable was borrowing *"couldn't reach the provider list"*, which sends a reader to check their network over a problem that is ours.

**2131 passing**, 0 warnings, app verified starting. Skips now 24, each logged with its id.

Deferred to #740 with the reviewer's agreement: the picker still resolves presets from the snapshot rather than the live list, costing a *"no API key stored"* hint for network-only presets. That's UI-side.

Issue #746, opened 2026-08-19
An About box: version, and credit where it is due


fsnow:
UI. Raised by **[fsnow]** while #738 was adding provider logos and needed somewhere to carry the models.dev MIT notice.

> **Provenance.** **[fsnow]** = binding. **[suggestion]** = advisory. **[measured]** = verified from the tree.

## What

**[fsnow]** "an About box with the CST Reader version as well as acknowledging models.dev and maybe some other open source projects".

There is no About surface today — **[measured]** no `AboutView`/`AboutWindow` anywhere in `src/CST.Avalonia/Views/`, and nothing in the menus. The version string exists only in the csproj and the packaging scripts.

## Why now

Two things want a home and neither has one:

- **The version.** A tester reporting a bug has no in-app way to say which build they are on. Beta 6 will have testers on macOS x64, macOS arm64, Windows x64 and Windows arm64, and the answer to "which one are you running?" is currently "check where you downloaded it from".
- **Attribution.** #738 caches MIT-licensed logo assets from models.dev and writes the licence into the cache directory beside them, which satisfies MIT but is not a place any reader will look. Several other components are worth naming for the same reason.

## What it needs to say

**[measured]** — the version, from the assembly rather than a hand-maintained string, so it cannot drift from what shipped:

| | |
|---|---|
| `Version` | `5.0.0-beta.6` |
| `AssemblyVersion` | `5.0.0.6` |
| `InformationalVersion` | `5.0.0-beta.6` |

**[suggestion]** platform and architecture alongside it, since that is the other half of "which build?" and `RuntimeInformation` already knows.

## The inventory

**[measured]** from the csproj files, so the issue starts from what is actually referenced rather than what anyone remembers:

**Texts and data** — the ones a reader would care about most:

- **Vipassana Research Institute** — the Chaṭṭha Saṅgāyana Tipiṭaka texts themselves. The app is a reader for VRI's corpus and that should be the first line, not a footnote among the libraries.
- **models.dev** (MIT) — the provider/model catalogue (#736) and the logos (#738).
- **Digital Pāḷi Dictionary** (CC BY-NC-SA) and **DPPN** — shipped as derived assets.
- **Childers, _A Dictionary of the Pali Language_ (1875)** — the bundled English dictionary.

[...]

PR #747, opened 2026-08-19
Fetch provider logos, keeping the monogram as the fallback (#738)

### LINKED ISSUE #738: Provider logos from models.dev, with the monogram as fallback

fsnow:
Closes #738. Last backend piece of #733; unblocks the logo half of #691.

## What

`AiProviderLogos` — fetch on demand per provider, cache to disk, and hand the UI a path or null. Null is the ordinary answer, not a failure: it means the monogram already on screen is what to draw.

## The part that is not obvious

**A missing logo returns 200, not 404.** models.dev serves a shared placeholder for any unknown id, so a status check would cache a grey stand-in for every provider without a mark. Measured byte-identical (1,421 bytes, sha256 `13bb0b37c627…`) across `definitely-not-a-provider`, `another-fake-xyz` and — the one that matters — `ollama`. It is recognised by content hash, and local runners keep their monogram exactly as #738 asked.

The hash is recorded as a pinned constant rather than probed at runtime, with a test pinning it. If upstream changes it the effect is a placeholder shown as though it were a logo — quiet rather than loud — which is why it is pinned.

## Two caching decisions

- **The placeholder is never written.** Caching it leaves a provider that gains a logo next month showing nothing until somebody clears the directory.
- **A failure is not remembered; a definite absence is.** A failed fetch says nothing about the provider, only about the moment. Remembering it would mean a reader who opened Settings before their wifi came up sees monograms for the rest of the session with no way to ask again. "This provider has no mark" is settled, so that one is remembered and stops asking.

## Where it caches

`~/Library/Application Support/CSTReader/cache/provider-logos/`, under `cache/` rather than beside `application-state.json` — these are purely derived, so deleting the directory should cost a refetch and nothing else. ~170 files at 355–1,421 bytes, fetched on demand rather than up front.

The MIT notice is written in beside the copies, which is where MIT asks for it. #746 (raised by @fsnow) tracks an About box, where a reader would actually see the acknowledgement.

## Hardening

Ids are slug-validated before use. A **connection** id is whatever its owner typed and it reaches the resolver as both a URL segment and a file name; the slug rule matches #737's and is what stops `../../..` from naming a file outside the cache.

## Binding surface

`LogoPath` and `HasLogo` on both row kinds, beside the existing `Monogram`/`MonogramTone`. Nothing here sizes, shapes or places a logo — that is the UI issues'.

A measured note that came out of writing the tests: **a connection added from the catalogue always carries the preset's id**, because `Save` routes preset adds through `AddFromPreset(_preset.Id, …)` and ignores the editor's id box. So those rows resolve reliably. A custom endpoint's id is no provider's; that row gets the monogram and nothing is guessed from it.

## ⚠️ The UI cannot consume this yet

**The app has no SVG renderer.** No `Avalonia.Svg.Skia`/`Svg.Skia` reference in any csproj, and Avalonia's `Image` takes a `Bitmap`. models.dev publishes only SVG, so displaying these means adding a rendering dependency — a dependency decision, so flagged rather than made here. Options and detail in https://github.com/fsnow/cst/issues/738#issuecomment-5346336212. The resolver degrades to the monogram on every path, so nothing is broken while this is undecided.

## Tests

21 new on the resolver, 6 on the rows. Full suite **2158 passed, 0 failed**.

The two placeholder tests were mutation-checked: with the detection removed they both fail. They needed it — a first draft that brute-forced the real hash passed while never finding it, because the resulting exception fell into the same `null` the fallback returns.

### DIFF
--- a/src/CST.Avalonia.Tests/Services/Ai/AiProviderLogosTests.cs
+++ b/src/CST.Avalonia.Tests/Services/Ai/AiProviderLogosTests.cs
@@ -1,5 +1,6 @@
 using System;
 using System.IO;
+using System.Linq;
 using System.Net;
 using System.Net.Http;
 using System.Security.Cryptography;
@@ -62,7 +63,8 @@ public class AiProviderLogosTests : IDisposable
         Assert.Equal(RealLogo, await File.ReadAllBytesAsync(first));
     }
 
-    /// <summary>Asked once per session, not per rebind — the connections list rebinds on every change.</summary>
+    /// <summary>Asked once per settled answer, not per rebind — the connections list rebinds on every
+    /// change.</summary>
     [Fact]
     public async Task A_second_request_does_not_hit_the_network()
     {
@@ -174,6 +176,8 @@ public class AiProviderLogosTests : IDisposable
     [InlineData("my/box")]
     [InlineData("my box")]
     [InlineData("-leading-dash")]
+    [InlineData("openai\n")]     // `$` would have matched before the newline, naming a file that contains one
+    [InlineData("OpenAI")]       // case-sensitive, like the id validators this mirrors
     public async Task An_id_that_is_not_a_slug_is_never_fetched(string id)
     {
         var stub = new Stub(() => Svg(RealLogo));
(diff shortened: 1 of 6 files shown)


fsnow:
Fable review done, all findings addressed. Rebased on main after #745.

## The one that mattered

**Any 200 that was not the placeholder was cached permanently as the logo.** `GetByteArrayAsync` throws on a non-2xx, so an error page with an error *status* could never get in — but a captive portal answers every request with 200 and a login page. Fable confirmed it empirically: a stub returning `<html><body>Sign in to Airport WiFi</body></html>` produced an `anthropic.svg` containing that HTML, and a fresh instance served it from disk with **zero** network calls.

One hotel wifi would have poisoned every visible provider until the reader found and deleted the cache directory — a transient network condition made into a permanent wrong answer, which is the precise failure this class refuses even to remember for a session. It also made the documented hash-drift mode worse than the comment claimed: a corrected hash in a later release could not heal a disk already poisoned, because the `File.Exists` hit never re-checked.

Fixed by sniffing the response for an SVG opening, and by re-reading and re-checking a cached file rather than trusting it — so a release can heal what an earlier one wrote. Mutation-checked: removing the two checks fails 5 tests.

## Concurrency was worse than the review estimated

Two rows can carry the same id — a preset row and the connection added from it — and load at the same moment. They shared a temp path, so the loser's `Move` found its own file gone and left that row on a monogram for the session, with `_asked` blocking any retry.

My first fix used the **process** id, which is not unique across threads of one process. The test I wrote caught it immediately: **7 of 8 concurrent callers got null.** Now unique per call.

## The rest

- **Slug rule was looser than the validators it claims to mirror.** `IgnoreCase` let `OpenAI` record an absence that `openai` then read — hiding a real provider's logo for the session — and admitted Unicode case-folds. `$` matches before a trailing newline, so `"openai\n"` passed and named a file containing a newline. Now case-sensitive, anchored `\z`.
- **Empty body** moved from "definite absence" to "transient". Fable was right that it is at least as plausibly a proxy artefact, and models.dev states a real absence with the placeholder.
- **A remembered path is re-checked against the disk**, since the cache directory is documented as safe to delete — otherwise rows bound to files that had gone, within the session.
- Timeout matched to `ModelsDevCatalog` rather than being a second number for one host; the interface no longer claims never to throw, since cancellation escapes by design.

## A vacuous test, again

`A_row_starts_on_its_monogram_before_anything_is_fetched` asserted a state that could not exist: the fake resolver answered synchronously, so the logo had already been applied before the constructor returned, and no mutation of the logo pipeline could fail it. Now holds the load open with a `TaskCompletionSource`, with a second test covering the swap. Same genus as the brute-force one I caught earlier — worth noting that two of the three tests I have written for this file needed a fake that does **not** complete synchronously to mean anything.

## Provenance

Fable flagged, correctly, that the session-memory policy — remember an absence, forget a failure — was **my** design and not in #738. Same for the slug validation, the empty-body handling and the notice placement. All are recorded in the commit messages as reasoning rather than as requirements, and none should be read as binding on #740.

**2184 passed, 0 failed.** App launched clean.

PR #749, opened 2026-08-19
Render cached provider logos with Avalonia.Svg.Skia (#748)

### LINKED ISSUE #748: Render a cached logo as geometry, so no SVG dependency is needed
Backend. Unblocks #740. Follows #738, which fetches and caches the SVGs but hands back bytes nothing can draw.

> **Provenance.** **[fsnow]** = binding. **[suggestion]** = advisory. **[measured]** = verified.

## The decision this rests on

**[fsnow]** "I would rather that the logos are dynamic, not compile time."

So the build-time route OpenCode uses — a vite plugin that pulls every logo into the bundle at `buildStart` (https://github.com/fsnow/cst/issues/740#issuecomment-5346722428) — is **not** what we do. #738's runtime fetch and disk cache stand as built. This issue supplies the missing half: turning a cached SVG into something Avalonia can draw, at runtime.

## Why no dependency is needed

**[measured]** `Avalonia.Media.Geometry.Parse` accepts SVG path mini-language directly. I extracted all 199 `d` strings from the 105 logos and parsed them in the running app: **`PARSED_OK=199 PARSED_FAIL=0`**.

Note for whoever tests this: `Geometry.Parse` **cannot run headless** — it throws `Unable to locate 'Avalonia.Platform.IPlatformRenderInterface'`. That is why this issue's deliverable stops at path-data strings and a matrix, and the single `Geometry.Parse` call belongs in the view. It keeps the whole parser testable in the suite.

## What the corpus actually contains

**[measured]** across the 105 logos that exist (of 192 providers — the rest have no mark and keep the monogram):

| Feature | Files | Handling |
|---|---|---|
| `<path>` | all | 199 elements; 64 files have exactly one, max 12 |
| `fill="currentColor"` only | 101 | → `Foreground`, so the mark themes itself light/dark |
| `stroke` | 24 | must be carried, or these render blank |
| `fill-rule` / `clip-rule` | 12 / 10 | evenodd → the `F0` prefix Avalonia's parser takes |
| `shape-rendering` | 15 | an antialiasing hint; ignorable, never geometry |
| `transform` | 5 | `scale()`, `translate()`, `rotate()`, `matrix()` — all standard, incl. **openrouter** |
| `opacity` / `fill-opacity` | 4 / 2 | carried |
| `<polygon>` / `<rect>` | 4 / 3 | convert to path data mechanically |
| `style="fill:…"` | 2 | 302ai, zenmux |
| `<clipPath>` + `<defs>` | 1 | novita-ai |
| no `viewBox` | 1 | dinference; has `width`/`height`, which is the default |

viewBoxes are not uniform — `0 0 40 40` (52), `0 0 24 24` (42), and outliers including `-14 -14 100.8 100.8`, so a **negative origin must be handled**, not assumed away.

## Deliverable

A parser from SVG bytes to a small drawable model — no Avalonia render types, so it tests headless:

- viewBox (falling back to `width`/`height`), including a negative origin
- per-shape: path data, fill (null meaning `currentColor`), fill-rule, stroke, stroke-width, opacity
- `polygon` and `rect` converted to path data
- `transform` parsed to a matrix and composed through nested `<g>`

**A whitelist, not a best effort.** Anything the parser does not fully understand returns null and the row keeps its monogram. **[suggestion]** a mangled logo is worse than a letter — the same reasoning that stops #738 guessing a provider from a renamed connection id.

**[measured]** expected coverage on today's corpus: ~104 of 105, with novita-ai (clipPath) the deliberate fallback. **[suggestion]** `style` parsing is 2 files and can go either way.

## Not this issue

Drawing them — size, placement, whether they appear in the picker as well as the list. That is #740.

fsnow:
Closes #748. Unblocks the logo half of #740.

## The decision

**[fsnow]** "Add the package" — after asking, correctly, whether I was writing an SVG parser. I was.

## What I had written, and why it went

A whitelist parser over an SVG subset — `path`/`polygon`/`rect`/`g`, viewBox, transforms, fill/stroke/fill-rule/opacity — so logos could be drawn as Avalonia geometry with **no dependency**. That part was real: **[measured]** `Geometry.Parse` takes SVG path data directly, and all 199 path strings in the current corpus parse (`PARSED_OK=199 PARSED_FAIL=0`, probed in the running app — it cannot run headless).

But the trade was bad, and worse *because* of **[fsnow]**'s "dynamic, not compile time" decision. With build-time assets I'd parse a set frozen at release and could see every file before shipping. Fetching at runtime, models.dev can publish a logo using a feature the subset misses at any moment, with no release of ours in between and nothing to signal it. The saving was one dependency; the cost was owning an SVG parser permanently. ~350 lines, deleted.

## The package

`Avalonia.Svg.Skia` **11.3.0**, MIT — same 11.3 line as our Avalonia 11.3.6. Binds SkiaSharp, which Avalonia already renders through, so it is a binding layer rather than a second engine. Brings `Svg.Skia`, `Svg.Model`, `Svg.Custom`, `ShimSkiaSharp`, `ExCSS`.

**[measured]** it draws every case the hand-rolled subset refused: `novita-ai`'s clipPath, `openrouter`'s transform, `dinference`'s missing viewBox, and the 15 files carrying `shape-rendering`.

**Worth stating:** its last release was 2025-05-04, ~15 months ago. That may only mean it tracks Avalonia 11.3 and is finished, but it is a maintenance signal rather than nothing.

## Theming had to be measured, not assumed

**[measured]** every logo renders `#000000` by default — including the **101 of 105** drawn in `currentColor`, which would be invisible on a dark ground.

The renderer accepts a stylesheet. Three candidates, rasterised and pixel-sampled in the app:

| Stylesheet | Result |
|---|---|
| `svg { color: … }` | **no effect** — the declaration does not inherit |
| `* { fill: … }` | **worse than nothing** — fills the outline-drawn logos solid |
| `* { color: … }` | **correct** — repaints `currentColor`, leaves brand colours alone |

Verified end to end: `currentColor` marks come back `#202020` against a light foreground and `#E8E8E8` against a dark one, while the four that name their own colours keep them.

**[measured]** exactly **4 of 105 will not theme** — `302ai`, `evroc`, `novita-ai`, `zenmux` — because they never mention `currentColor`. Two are black, so they will be dim on a dark ground. **[suggestion]** that is #740's to judge with the app in front of it; a tile behind the mark, as the monogram already has, would settle it.

## Testing

**No unit tests, deliberately.** `SvgSource.Load` needs a render backend — the same reason `Geometry.Parse` throws `Unable to locate 'Avalonia.Platform.IPlatformRenderInterface'` headless. Verified in the running app instead, by rendering to a `RenderTargetBitmap` and counting and sampling pixels; the null paths (missing file, null path) were checked the same way.

Full suite **2184 passed, 0 failed**. App launched clean.

## Not committed

The 105 vendor SVGs. I had staged them as test fixtures and removed them: putting the asset set in the repo is close to the compile-time shape **[fsnow]** rejected, even in a test project.

### DIFF
--- a/src/CST.Avalonia.Tests/Services/Ai/AiLogoImagesTests.cs
+++ b/src/CST.Avalonia.Tests/Services/Ai/AiLogoImagesTests.cs
@@ -1,0 +1,145 @@
+using System;
+using System.IO;
+using System.Text;
+using Avalonia.Media;
+using CST.Avalonia.Services.Ai;
+using Xunit;
+
+namespace CST.Avalonia.Tests.Services.Ai;
+
+/// <summary>
+/// #748: what may be handed to the SVG renderer.
+///
+/// <para>Rendering itself is not covered here: it needs Avalonia's rendering stack, and building an
+/// <c>SvgImage</c> is thread-affine. The screening in front of it is plain file-and-string work, and is where
+/// the consequential decisions live, so that is what these cover. The render path was verified in the running
+/// app by rasterising to a bitmap and sampling pixels; see the PR.</para>
+///
+/// <para><b>The thread guard is not covered here, and cannot be.</b> With no Avalonia application running,
+/// <c>Dispatcher.UIThread</c> binds to whichever thread first touches it — in a test process that is a pool
+/// thread, so <c>CheckAccess()</c> returns true off-thread and the guard never trips. A test asserting it
+/// would pass whether or not the guard existed.</para>
+///
+/// <para>Why screening exists at all: these files arrive over the network. The renderer expands XML entities
+/// with no quota, so a few KB of nested declarations become gigabytes, and it fetches
+/// <c>&lt;image href="https://…"&gt;</c> synchronously on the calling thread — which must be the UI thread.
+/// Neither appears in any of today's logos. (fable review)</para>
+/// </summary>
+public class AiLogoImagesTests : IDisposable
+{
+    private readonly string _dir = Path.Combine(Path.GetTempPath(), "cst-748-" + Guid.NewGuid().ToString("N"));
+
+    public AiLogoImagesTests() => Directory.CreateDirectory(_dir);
+
+    public void Dispose() { try { Directory.Delete(_dir, true); } catch { } }
+
+    private string Write(string name, string content)
+    {
+        var path = Path.Combine(_dir, name);
+        File.WriteAllText(path, content);
+        return path;
+    }
+
+    /// <summary>The shape of a real logo: an SVG namespace whose value is itself an http URL, which must not
+    /// read as "references a remote resource".</summary>
+    [Fact]
+    public void A_real_logo_passes()
+    {
+        var path = Write("anthropic.svg", """
+            <svg width="24" height="24" viewBox="0 0 40 40" xmlns="http://www.w3.org/2000/svg">
+            <path d="M26.9 9.8H22.1L30.7 31.7H35.4Z" fill="currentColor"/></svg>
+            """);
+
+        Assert.Null(AiLogoImages.Screen(path));
+    }
+
+    /// <summary>Billion laughs. A few KB of nested declarations expand to gigabytes, and the renderer sets no
+    /// quota — so the cost is a frozen window long before the failure.</summary>
+    [Fact]
+    public void A_document_declaring_entities_is_refused()
+    {
+        var path = Write("bomb.svg", """
+            <?xml version="1.0"?>
+            <!DOCTYPE svg [<!ENTITY a "aaaaaaaaaa"><!ENTITY b "&a;&a;&a;&a;&a;&a;&a;&a;&a;&a;">]>
+            <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 10 10"><path d="&b;"/></svg>
+            """);
+
+        Assert.Equal("declares XML entities", AiLogoImages.Screen(path));
+    }
+
+    /// <summary>Fetched synchronously, on the UI thread, with the default 100-second timeout. A decorative
+    /// icon has no business reaching the network at draw time.</summary>
+    [Theory]
+    [InlineData("""<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 10 10"><image href="https://evil.example/x.png"/></svg>""")]
+    [InlineData("""<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 10 10"><image xlink:href="http://evil.example/x.png"/></svg>""")]
+    [InlineData("""<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 10 10"><image href="file:///etc/passwd"/></svg>""")]
+    [InlineData("""<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 10 10"><path fill="url(https://evil.example/g)" d="M0,0H1"/></svg>""")]
+    public void A_document_reaching_off_the_machine_is_refused(string svg)
+    {
+        Assert.Equal("references a remote resource", AiLogoImages.Screen(Write("remote.svg", svg)));
+    }
+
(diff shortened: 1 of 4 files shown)


fsnow:
Fable review done. It found one thing I had missed completely, and it changes the package version.

## The app's Skia engine was being swapped out from under Avalonia

`Avalonia.Svg.Skia` **11.3.0** → `Svg.Skia` 3.0.2 → demands `SkiaSharp >= 3.116.1`. Avalonia.Skia 11.3.6 declares **2.88.9**, so NuGet unified the whole graph upward. **Every glyph Avalonia draws — all 14 scripts — would have rendered through a Skia one major version newer than Avalonia declares**, as an unstated side effect of a logo feature.

I verified it rather than taking it on report, by diffing `project.assets.json` across the two branches:

| | main | branch (11.3.0) |
|---|---|---|
| `SkiaSharp` | 2.88.9 | **3.116.1** |
| `NativeAssets.macOS` / `Win32` | 2.88.9 | **3.116.1** |
| `NativeAssets.Linux` | 2.88.9 | **2.88.9** ← left behind |

That last row is its own tripwire: a future `linux-x64` publish would pair managed SkiaSharp 3.116.1 with native 2.88, which fails SkiaSharp's version check at startup.

I wrote "a binding layer rather than a second engine" in the PR body. That was wrong, and it was wrong in the direction that sounds reassuring.

## The fix: 11.2.0.2 instead

`Avalonia.Svg.Skia` **11.2.0.2** depends on `SkiaSharp` **2.88.8**, which unifies harmlessly with Avalonia's own 2.88.9. **[measured]** resolution is now byte-identical to main — Linux natives included — so the engine question disappears rather than being accepted.

**[measured]** it renders every case identically to 11.3.0, theming included: `novita-ai`'s clipPath, `openrouter`'s transform, `dinference`'s missing viewBox, `azure-cognitive-services`' shape-rendering, and `currentColor` recolouring to `#202020`/`#E8E8E8` on both grounds.

## Hardening the renderer's input

Two things the renderer does that a decorative icon has no business doing, both traced by Fable into the dependency source:

- **XML entities expand with no quota.** A few KB of nested declarations become gigabytes. It passes #738's SVG sniff, since `<?xml` and `<!DOCTYPE svg` are valid openers.
- **`<image href="https://…">` is fetched synchronously**, on the calling thread — which must be the UI thread — with `WebRequest`'s 100-second default timeout. One reference is one frozen window, per theme colour.

Neither appears in today's 105 logos. But these files come off the network, which is exactly why the set is not ours to assume. Files are now screened on size, entity declarations, and any absolute URL once the SVG namespace is discounted.

## Failures are no longer cached

The previous version stored a null forever for that (path, colour) — contradicting the policy #738 states at length for itself, with a concrete trigger: #738 deletes and refetches to the *same path* when healing a poisoned entry, so a render inside that window left the row on a monogram for the session. Only successes are cached now.

`Get` also returns null off the UI thread rather than risking a throw inside a fire-and-forget binding, and does not remember that null either.

Alpha is now dropped consistently — it was already dropped from the render but kept in the cache key, so colours differing only in alpha made duplicate entries with identical output.

## Two corrections to my own record

- **`Svg.Custom` is MS-PL, not MIT.** Permissive and compatible, but I said MIT and that was wrong.
- **"Cannot be unit tested" was overstated.** Only `SvgImage` construction is thread-affine; the SVG parse itself touches no Avalonia platform service. The screening is now covered by **13 tests**, and its wiring into the render path was checked in the app: a billion-laughs file and a remote-reference file are both refused and logged while a real logo renders.

One gap stated plainly: the tests exercise `Screen`'s logic, not that `Render` calls it — that wiring needs the render path, so it is covered by the in-app check above rather than by the suite. And the thread guard cannot be covered at all: with no Avalonia app running, `Dispatcher.UIThread` binds to whichever thread first touches it, so `CheckAccess()` returns true off-thread and a test would pass whether or not the guard existed.

**2197 passed, 0 failed.** App launched clean.

Issue #678, opened 2026-08-17
Credentials are keyed by provider kind, so two OpenAI-compatible providers collide


fsnow:
`AiCredentialStore.AccountFor` maps a `ChatProviderKind` to a Keychain account: `anthropic` or `openai-compatible`. So **every OpenAI-compatible endpoint shares one credential slot**.

OpenRouter, DeepSeek, Together, Fireworks, a hosted vLLM and a local Ollama are all `openai-compatible`. A reader who stores an OpenRouter key and then switches to DeepSeek silently sends the wrong key, and gets a 401 that names neither cause.

## Why it matters now

It blocks cross-model testing, which is the point of the eval work — you cannot hold keys for two hosted providers at once. It also makes the model picker (#674) confusing: switching to an id from another provider needs that provider's base URL *and* its key, and today there is only one key.

## Shape

Key on the **endpoint**, not the kind. The base URL is already what selects the provider for the OpenAI-compatible adapter — the code says so explicitly — so it is the natural account key. Normalised host, or host plus path prefix.

Care needed on:

- **Migration.** An existing single `openai-compatible` entry has no recorded endpoint. Do not silently re-key it to whatever is configured now; that would attach a key to an endpoint the reader never associated it with.
- **Reporting.** "No key stored for this endpoint" must name the endpoint, since the whole failure mode is a key that exists but belongs to a different one.
- **Windows.** #579's DPAPI half is still outstanding and would inherit whatever shape this takes.

Related: #579, #609, #674.

fsnow:
## Prior art: how OpenCode Desktop models a connection

Studied OpenCode Desktop v1.18.18's settings (8 screenshots + one live failure test, 2026-08-17) purely as prior art for this issue and #674. It has already made most of the calls this issue is asking for, and one of them is better than what this issue currently proposes.

### A connection is a record with a stable slug, not a URL

Their Custom-provider sheet asks for, in order:

| Field | Notes |
|---|---|
| **Provider ID** | first/focused field, placeholder `myprovider`, helper text "Lowercase letters, numbers, hyphens, or underscores" |
| **Display name** | placeholder `My AI Provider` |
| **Base URL** | placeholder `https://api.myprovider.com/v1` |
| **API key** | *"Optional. Leave empty if you manage auth via headers"* |
| **Models[]** | repeating `model-id` + `Display Name` rows, trash icon, `+ Add model` |
| **Headers[]** | optional, same repeating-row idiom |
[...]

PR #760, opened 2026-08-20
Stop calling a required API key optional (#691)

fsnow:
A named provider's sheet — DeepSeek, OpenRouter, anything with a key box — showed this under the box:

> Optional — leave it empty if this endpoint needs no key, or if you authenticate with a header below.

Both halves are wrong there. The blurb three lines above says *"Enter your DeepSeek API key to use DeepSeek models in CST Reader"*, so the sheet contradicts itself and invites the reader to save a connection that cannot answer. And the header it points at is never shown: headers are asked for on a custom endpoint alone.

The two ways to reach that sheet are exhaustive — a provider needing no key shows no key box at all (`ShowKeyField`) — so any box on a preset sheet is a required one. The line now shows for a custom endpoint alone, where the box genuinely is optional and the headers exist.

Two tests: a required key is not called optional, and a custom endpoint keeps the line together with the headers section it refers to.

### DIFF
--- a/src/CST.Avalonia.Tests/ViewModels/AiConnectionEditorViewModelTests.cs
+++ b/src/CST.Avalonia.Tests/ViewModels/AiConnectionEditorViewModelTests.cs
@@ -372,6 +372,41 @@ public class AiConnectionEditorViewModelTests
         Assert.False(vm.HasKeyHint);
     }
 
+    /// <summary>
+    /// Editing a connection added from a key-requiring provider does not call the key optional either.
+    ///
+    /// <para>Gating on "is this a preset sheet" would leave the line on the sheet a reader is most likely to
+    /// be reading it on: Edit is reached precisely because the key is missing or wrong, one click from where
+    /// the contradiction was reported. The edit form carries no preset, so the requirement has to be
+    /// recovered from the id — which is unambiguous, a custom connection being refused a preset's id.
+    /// (fable review)</para>
+    /// </summary>
+    [Fact]
+    public void Editing_a_named_provider_does_not_call_its_key_optional()
+    {
+        var h = new Harness();
+        h.Service.AddFromPreset("deepseek", new Dictionary<string, string>());
+
+        var vm = h.Existing("deepseek");
+
+        Assert.True(vm.ShowKeyField);
+        Assert.False(vm.HasKeyHint);
+    }
+
+    /// <summary>Editing a custom endpoint keeps the line: nothing about it requires a key.</summary>
+    [Fact]
+    public void Editing_a_custom_endpoint_keeps_the_optional_line()
+    {
+        var h = new Harness();
+        Save(h.Custom().With(vm =>
+        {
+            vm.Id = "my-box";
+            vm.BaseUrl = "http://localhost:1234/v1";
+        }));
+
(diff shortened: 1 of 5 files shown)


fsnow:
Fable reviewed this and found the gate too narrow, pushed as `fc717d4`.

**The case split I called exhaustive wasn't.** There are three factory methods, not two — `ForExisting` also passes `preset: null`, so Edit mode kept the line, including for a connection added from a key-requiring provider. And that is the sheet a reader is most likely to be reading it on: add DeepSeek without a key, watch the models grey out with "no API key stored", click Edit to paste one, and the sheet again says the box is optional. One click from where the contradiction was reported.

The requirement is now recovered from the connection's id. That lookup is exact rather than a guess: `ValidateId` refuses a custom connection any preset's id outright (*"'deepseek' is the id of a built-in provider"*), so an id that matches one came from it. Deliberately not stored as `_preset` — that field decides which *fields* the sheet shows, and an edit must keep showing all of them.

Two more tests: editing a named provider doesn't call its key optional, editing a custom endpoint keeps the line.

Three attacks Fable ran that didn't land, verified against the code rather than assumed:

- **The env-var escape.** `RequiresKey` is true for `Env` methods too, so a key supplied by `DEEPSEEK_API_KEY` would make the box genuinely optional and the removed line the only thing saying so. But the app never reads those variables — `Environment.GetEnvironmentVariable` appears only in `SingleInstanceGuard` and unrelated startup code, and `SourceFor` only ever returns `Keychain` or `None`. `CredentialSource.Environment` is plumbing for a feature that doesn't exist yet.
- **A blank space under the box.** `KeyStatus` has no visibility gate, so a preset sheet still shows the credential-store line.
- **The tests passing for the wrong reason.** `Harness.Preset` uses `.Single(...)`, which throws rather than yielding a null preset; "deepseek" is real, from the models.dev snapshot.

One pre-existing gap noted, not touched here: saving a preset connection with an empty key isn't blocked — `AddFromPreset` validates only URL placeholders — so you can create a connection that 401s. The picker catches it downstream with "no API key stored". Say the word if you want that filed.

fsnow:
Filed as #761 and fixed here in `d144e26`, since it is the same sheet and the same sentence — the line said the key was optional, and the save agreed with it.

Pressing **Add** with the key box empty now stops at the sheet's existing `Problem` line, where the service's other refusals already appear. Not where the credential store is unavailable: that sheet already says so in caution colour and no key can be filed by any route there, so a refusal on top would leave the reader nowhere to go.

Ordering note: where a preset needs both an input and a key — Azure with no resource name and no key — the key is refused first, then the resource name on the next press. Two presses in a case that needs both, rather than duplicating the service's own placeholder validation in the view model to merge the messages.

**Two existing tests were passing by adding a keyless OpenRouter**, so this was load-bearing in the suite as well as in the app. They now supply a key, which is how a reader adds one. `AddThroughSheet` supplies one by default and types it only where the sheet shows the box, so a local runner still stores nothing — without that guard it filed a key under Ollama and broke the test that proves keys land under the right connection (#678's hazard).

Five new tests: the refusal, the same sheet saving with a key, an edit not demanding a key that is already stored, and a local runner and a custom endpoint both still adding with nothing typed.

Full suite: 2250 pass, 0 build warnings.

Issue #579, opened 2026-08-09
Surface B: credential storage — Keychain (macOS) + DPAPI (Windows)


fsnow:
Part of #186 (surface **B**). Plan: [`docs/features/planned/AI_SURFACE_B.md`](https://github.com/fsnow/cst/blob/main/docs/features/planned/AI_SURFACE_B.md) §6.

Greenfield — the app stores no secrets today. (`use-mock-keychain` in `App.axaml.cs` scopes Chromium's OSCrypt; it does not constrain our process.)

- **macOS** — Keychain via Security.framework P/Invoke, one generic-password item per provider.
- **Windows** — DPAPI (`ProtectedData`, CurrentUser scope) into the app data directory.
- **Never in `settings.json`** — that file is hand-edited, screenshotted, and pasted into bug reports.
- **Read lazily**, on first request, not at startup — so a user who never enables surface B never triggers a Keychain prompt. Same class of mistake as the Chromium Safe Storage prompt we already work around.
- **Never logged at any level.**

### Acceptance

A test asserting the key string never appears in log output. The DPAPI path needs a Windows target (Merlin or Placid) — it cannot run under `dotnet test` on the macOS dev machines.

**Linux is deliberately unaddressed** while it is unshipped, but define the behaviour (surface B unavailable, or key-per-session) so it is not an accidental gap.

**Depends on:** nothing. **Platform-gated, not fully UI-free.**

fsnow:
**The macOS half landed** in #608 (`ab3c8a2`). Left open for the Windows DPAPI half.

Done:

- Common seam — `IAiCredentialStore` (get / set / delete / availability), wired into the provider resolver and DI.
- macOS Keychain via `SecItem*` P/Invoke, with the `kSec*` constants read by `dlsym`. Round trips verified against the real Keychain.
- The acceptance test: the key never appears in log output at any level, across store/read/delete, and against the **structured** log values as well as the formatted message.
- **Linux behaviour defined** rather than left as an accidental gap: reports unavailable with a reason, and says a keyless endpoint still works — the local-runner path stays open there.
- `settings.json` never carries a key.
- Read lazily, on the request that needs it; nothing at startup, so a user who never enables surface B never triggers a Keychain access.

Remaining for this issue:

[...]

PR #764, opened 2026-08-21
Send a connection's headers, on both provider kinds (#711)

### LINKED ISSUE #711: Connection headers are collected and stored, but never sent with a chat request
A connection carries `Headers[]`, the editor collects them, and they are stored in `settings.json` — but
nothing sends them. `ChatProviderResolver` builds `OpenAiCompatibleOptions(baseUrl, apiKey)` and
`AnthropicOptions(apiKey, baseUrl)`, neither of which has a headers field, and neither provider adds any
beyond the credential:

- `OpenAiCompatibleProvider.cs:74` sets `Authorization: Bearer` and nothing else.
- `AnthropicMessagesProvider.cs:75-76` sets `x-api-key` and `anthropic-version` and nothing else.

## Why it matters

`Headers[]` is not a nicety — it is the mechanism that makes an **absent API key coherent**. #689 introduced
it as "the escape hatch": *"leave empty if you manage auth via headers"* covers Azure's `api-key`, gateway
tokens, and anything else that is not a bearer credential. The connection editor says exactly that to the
reader. So the field that authenticates a whole class of endpoints is collected, persisted, displayed back —
and dropped on the floor at request time.

The failure is silent and misattributed. A reader configures a gateway token as a header, sends a question,
and gets a 401 that names nothing. Everything on the settings screen says the connection is set up correctly,
because as far as the settings screen is concerned it is.

## It is about to get worse, not better

The model-catalogue fetch in #674 **does** apply `Headers[]` (`AiModelCatalog.Authenticate`), because it had
to in order to reach Azure. So on the same connection:

- the Models tab lists the provider's models successfully, and
- asking a question fails with a 401.

Two surfaces disagreeing about whether a connection works is precisely the failure #673 was about, arriving
from the opposite direction. A reader would reasonably conclude the model list proves the credentials are
fine and look for the problem somewhere else entirely.

## Shape

Carry the headers to the request:

- add a headers field to `OpenAiCompatibleOptions` and `AnthropicOptions`,
- have `ChatProviderResolver` pass `connection.Headers` (expanded through `AiTemplate` with the connection's
  `Inputs`, as the catalogue fetch does — Azure's are templated),
- apply them last in both providers, **replacing** rather than adding to the credential header where they
  collide. Azure authenticates with `api-key` and expects `Authorization` to be *absent*, not also present,
  so an additive merge fails there.

`AiModelCatalog.Authenticate` already does exactly this and is worth reading as the reference — ideally the
two share one helper rather than growing a second opinion about how a connection authenticates.

## Test worth having

A connection with `Headers = { "X-Gateway": "token" }` and no API key should produce a chat request carrying
that header. There is currently no test that would fail if headers were dropped, which is how this got this
far.

Found while implementing #674. Backend, so filing rather than fixing — the UI half already collects and
displays them correctly.

fsnow:
Closes #711.

## What was broken

`Headers[]` were collected by the editor and written to `settings.json`. Two separate things then dropped them.

**The Anthropic adapter had nowhere to put them.** `AnthropicOptions` was `(ApiKey, BaseUrl)` — the resolver could not have passed headers if it had wanted to. That is the case the field exists for: `Kind = Anthropic` does not imply `api.anthropic.com`, and a gateway speaking the Messages protocol at its own URL is exactly where a second header lives.

**`AiHttp.ApplyAuth` skipped any extra header named `Authorization`, unconditionally.** The guard is meant to stop a mistyped header from overwriting the credential. It fired in two cases where there was no credential to protect:

- a custom auth header is configured (Azure's `api-key`) and the reader deliberately adds `Authorization` for a gateway in front of it;
- **no key is stored at all** — the "leave the key empty if you manage auth via headers" escape hatch #689 promised, defeated by the header name a gateway most often wants.

The guard now covers the credential header, and only when a credential is actually being attached.

## The part worth a second opinion

Relaxing that exposed a third guard. **Both** the resolver and the Anthropic adapter refused a keyless Anthropic connection outright, so headers-only authentication would still have failed — the fix would have been inert in the one scenario it exists for. Both now refuse only a connection that has *neither* a key *nor* a header.

That is a deliberate widening of "Anthropic requires a key" to "Anthropic requires a key or headers", and it is the one judgement call here rather than a mechanical fix. The ordinary keyless case still gets the same message it always did, including the `_credentials.Unavailable` branch from #579.

`anthropic-version` is added only when the connection did not supply one. `TryAddWithoutValidation` appends rather than replaces, so adding ours unconditionally to a request that already carried one would send `anthropic-version: 2026-01-01, 2023-06-01` — two values, worse than either. A gateway that pins its own version gets it.

## Verification

7 new tests. The Anthropic ones assert at the wire (`StubHttpMessageHandler.LastRequest`) rather than through the header-building method, because the defect they replace was a request shape that read correctly in the code and arrived incomplete.

Mutation-checked in both directions:

- reverting the adapter to `extraHeaders: null` → 3 of the 5 `AnthropicAuthTests` fail
- restoring the blanket `Authorization` skip → both new `OpenAiCompatibleAuthTests` fail

Full suite: **2267 passed, 0 failed, 17 skipped** (up from 2260; the skips are the Windows DPAPI and instance-activation tests, platform-gated).

Not covered: no live gateway was exercised. The headers are asserted onto the outgoing request, not through a real Cloudflare or Azure endpoint.


fsnow:
Fable reviewed this adversarially and found four confirmed defects plus two minor ones. All are now fixed in dcf0d9f. Summary, since the review changes what the PR body claims:

**1 — the model-listing path contradicted the chat path (CONFIRMED).** `AiModelCatalog.Authenticate` added `anthropic-version` unconditionally; the adapter added it only when absent. A connection pinning its own version therefore sent one value on a chat request and **two** on a model listing. That is exactly the "list loads while answers fail" divergence the shared auth helper's doc comment asserts cannot happen — arriving through the one header the helper does not own. **The claim in this PR's body that `anthropic-version` can never be duplicated was false for that path.**

**2 — the widened guard was satisfiable by nothing (CONFIRMED).** `HasCredential` counted any header with a non-blank *name*. A cosmetic `X-Title`, an empty value, or a name HTTP rejects outright all counted, so a keyless connection passed both guards, sent nothing that could authenticate, and the resulting 401 surfaced as *"The provider rejected the API key"* — naming a key that does not exist, and replacing a refusal that named the actual fix. It now counts only a header that will reach the wire carrying something: legal field-name token, non-blank value, no unresolved placeholder.

It remains a proxy, deliberately. We cannot know which header a gateway treats as its credential, and guessing by name would be curation in a different costume. It is now only a proxy that cannot be satisfied by nothing.

**3 — the 401 message (CONFIRMED, follow-on).** A 401 on a connection with no stored key now says so instead of blaming the key.

**4 — unfilled header templates (PLAUSIBLE, fixed).** A header value with an unresolved `{placeholder}` reached the wire verbatim, though the base URL beside it has been refused for exactly this since #689. In the gateway case the header *is* the credential, so an unfinished one is an unfinished credential. Now refused, naming the missing input.

**5 — orphaned doc comment (minor, fixed).** The `ApplyHeaders` summary had landed on `HasCredential`.

**6 — CR/LF in a header value (minor, not fixed here).** Confirmed pre-existing on the OpenAI-compatible path: a pasted value with a stray newline is accepted into the request and fails at send as a transport exception misreported as a connectivity problem. Not header injection — .NET refuses to serialize it. Out of scope for this PR; worth its own issue if it is worth fixing at all.

Findings 1–3 are mutation-checked: reverting either guard fails 6 tests. Full suite **2279 passed, 0 failed, 17 skipped** (up from 2260 on main).

**Verdict on the widening**, which is what I flagged for a second opinion: the direction was judged correct and necessary — without it the headers fix is inert in the one Anthropic case it exists for — but the predicate as first written was the weakest possible one, and it shipped without repairing the message it invalidated. Both repairs Fable asked for are in.

**Separately, not a defect in this PR:** `AiConnectionsViewModelTests.RepoRoot()` walks up looking for a `.git` **directory**. In a git worktree `.git` is a file, so those tests fail in any agent worktree while passing in the main checkout. That silently penalises every worktree-isolated review.

PR #768, opened 2026-08-21
Measure whether an endpoint takes /v1, instead of assuming (#742)

### LINKED ISSUE #742: A bare-host base URL always gains /v1, which is wrong for at least one real provider

fsnow:
Closes #742. **Stacked on #764** — base branch is `fix/711-anthropic-extra-headers`, because both touch `AiHttp`, `AiModelCatalog` and `ChatProviderResolver`. Merge #764 first and this retargets to main cleanly.

## The measurement that changed the plan

The issue leaned toward putting the convention on the connection and letting a preset set it. Checking the catalogue first says that would not have reached the reported bug:

- **models.dev carries no `api` URL for Perplexity at all** — which is exactly why #737 skipped it. A preset can never carry its convention because there is no preset.
- Of the 192 providers in the snapshot, **166 publish an `api` URL: 157 already carry a version segment, 7 have some other path, and exactly 2 are bare hosts** (DeepSeek, GitHub Copilot) — both of which do want `/v1`.

So the bare-host rule almost never fires for a preset. It fires for **hand-typed** URLs, which is precisely where there is no preset to consult. Hence: learn it.

## What it does

When a model listing 404s against a URL we guessed at, it asks the other way **once**. Whichever answers is recorded on the connection as `UsesVersionSegment`, and every later request — chat included — obeys it.

```
Add connection  →  GET https://api.perplexity.ai/v1/models   404
                →  GET https://api.perplexity.ai/models      200
                →  record UsesVersionSegment = false

Every later request:  POST https://api.perplexity.ai/chat/completions
```

One extra request, once per connection, at a moment the reader is already waiting on a listing. Nothing on the answer path pays for it.

## What it deliberately does not do

- **No retry on the chat path.** 404 is also what several OpenAI-shaped providers return for an unknown *model id*, so a retry there fires on a failure it cannot fix and makes a genuinely wrong URL slower to diagnose.
- **No editor field.** Asking readers whether their provider versions its path is asking about the one thing this bug hides from them. It is measured, not configured.
- **No second-guessing a URL the reader fully specified.** A base that already names its version was not a guess, so its 404 is the endpoint's answer and costs exactly one request.
- **A recorded answer replaces a guess, never a fact about the URL in hand.** A base carrying `/v1` keeps it whatever was recorded, so a stale record cannot start mangling a correct URL. Pinned by a `[Theory]` over both recorded values.

Retyping the base URL forgets the measurement — it is a fact about one host, and carrying it to a new one is the same silent-rewrite failure in reverse. Renaming the connection keeps it.

## Verification

11 new tests across two files. `ReportEndpointVersioning` writes only on change, so the Models tab recording this on every successful fetch does not churn settings or `ConnectionsChanged`.

One bug my own tests caught before review: `Apply` overwrote `record.BaseUrl` *before* my comparison against it, so the forget-on-retype reset never fired. Moved above the assignment.

Full suite: **2290 passed, 0 failed, 17 skipped**.

Not covered: no live Perplexity endpoint was exercised. The counterexample is the vendor SDK measurement recorded in the issue; the probe is asserted against a stub.

### DIFF
--- a/src/CST.Avalonia.Tests/Services/Ai/AiBaseUrlVersionTests.cs
+++ b/src/CST.Avalonia.Tests/Services/Ai/AiBaseUrlVersionTests.cs
@@ -1,0 +1,150 @@
+using System.Collections.Generic;
+using System.Net;
+using System.Net.Http;
+using System.Threading.Tasks;
+using CST.Avalonia.Models.Ai;
+using CST.Avalonia.Services.Ai;
+using CST.Avalonia.Tests.TestSupport;
+using Microsoft.Extensions.Logging.Abstractions;
+using Xunit;
+
+namespace CST.Avalonia.Tests.Services.Ai;
+
+/// <summary>
+/// #742: a bare-host base URL always gained a <c>/v1</c>, which is wrong for at least one real provider.
+/// Perplexity documents <c>https://api.perplexity.ai</c> and serves <c>/chat/completions</c>, so every
+/// request 404'd against a URL the reader could not see us rewriting.
+///
+/// <para>The answer is measured, not configured: the model listing asks once, and what it learns is recorded
+/// on the connection so no later request has to guess. These tests pin the three parts of that — the guess,
+/// the probe, and the memory.</para>
+/// </summary>
+public class AiBaseUrlVersionTests
+{
+    private static AiConnection Connection(string baseUrl, bool? usesVersionSegment = null) => new(
+        Id: "custom",
+        DisplayName: "Custom",
+        Kind: ChatProviderKind.OpenAiCompatible,
+        BaseUrl: baseUrl,
+        Models: new List<AiModelEntry>(),
+        Headers: new Dictionary<string, string>(),
+        Inputs: new Dictionary<string, string>(),
+        UsesVersionSegment: usesVersionSegment);
+
+    private static (AiModelCatalog Catalog, List<string> Urls) Catalog(
+        params (HttpStatusCode Status, string Body)[] responses)
+    {
+        var urls = new List<string>();
+        var next = 0;
+        var handler = new StubHttpMessageHandler(request =>
+        {
+            urls.Add(request.RequestUri!.ToString());
+            var (status, body) = responses[next < responses.Length ? next : responses.Length - 1];
+            next++;
+            return new HttpResponseMessage(status) { Content = new StringContent(body) };
+        });
+
+        return (new AiModelCatalog(new HttpClient(handler), null, NullLogger<AiModelCatalog>.Instance), urls);
+    }
+
+    private const string OneModel = """{"data":[{"id":"sonar-pro"}]}""";
+
+    // ---- the guess, unchanged where it was right ---------------------------------------------------------
+
+    [Fact]
+    public async Task A_bare_host_is_still_tried_with_the_version_segment_first()
+    {
+        var (catalog, urls) = Catalog((HttpStatusCode.OK, OneModel));
+
+        var result = await catalog.FetchAsync(Connection("https://api.deepseek.com"));
+
+        Assert.True(result.Ok);
+        Assert.Equal("https://api.deepseek.com/v1/models", Assert.Single(urls));
+        // Nothing was learned, because nothing needed to be: the guess was right on the first try.
+        Assert.Null(result.UsesVersionSegment);
+    }
+
+    // ---- the probe ---------------------------------------------------------------------------------------
+
+    [Fact]
+    public async Task A_404_on_the_guessed_url_asks_the_other_way_once_and_reports_what_worked()
+    {
+        var (catalog, urls) = Catalog(
+            (HttpStatusCode.NotFound, "not found"),
+            (HttpStatusCode.OK, OneModel));
+
+        var result = await catalog.FetchAsync(Connection("https://api.perplexity.ai"));
+
+        Assert.True(result.Ok);
+        Assert.Equal(
+            new[] { "https://api.perplexity.ai/v1/models", "https://api.perplexity.ai/models" },
+            urls);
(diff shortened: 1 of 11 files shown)


fsnow:
Converted to draft. Fable's verdict is **do not merge as-is**, and the central finding invalidates the design rather than the implementation. Recording it here because the measurement is the valuable part.

## The design is wrong, and the motivating provider is what proves it

Fable measured `api.perplexity.ai` live on 2026-08-21:

| request | result |
|---|---|
| `GET /v1/models` | **401** — route exists |
| `GET /models` | 404 |
| `POST /chat/completions` | **401** — route exists |
| `POST /v1/chat/completions` | 404 |

**Perplexity's listing is versioned while its chat is not.** One host, two conventions. The issue hinted at exactly this and I read past it — it records `baseURL}/chat/completions` alongside `baseURL}/v1/embeddings` and even notes "its own embeddings route does, which is presumably how the convention came to differ."

This PR measures the *listing* route and stamps the answer on the *chat* route. For a real Perplexity user with a valid key, `/v1/models` succeeds on the first try, nothing is learned, and every chat turn still 404s. In the branch where the listing does 404, the alternative `/models` 404s too — so still nothing is learned. **Both paths leave the filed bug exactly where it was**, while the now-successful model listing tells the reader the connection works. That is the two-surfaces contradiction #673 exists to prevent, introduced by a fix for something else.

## And it can make things worse

For any provider with a readable `/models` and chat at `/v1/chat/completions`, the probe records `UsesVersionSegment = false` and flips a **working** chat path to 404. The record is then effectively unfixable: the probe never re-fires, nothing in the UI shows the record exists, and the only reset is typing a *different* URL — so re-pasting the same documented URL, the natural repair, clears nothing.

That answers the question I put in the PR body. A wrong record is worse than the original bug: the original 404'd one provider shape from the start, this can silently break a connection that was working.

## Also confirmed, independent of the redesign

- **A failed retry misattributes both URL and reachability.** If the retry throws or times out, the catch blocks still name the original URL — which *answered* — and report `reachable: false`, marking a connection Unreachable moments after it responded 404. Contradicts `AiCatalogResult.Reachable`'s own contract.
- **In-flight fetch vs. edit race** defeats forget-on-retype: a fetch started before an edit completes after it, stamping a measurement of the old host onto the new URL. Creating the #742 failure class with the #742 fix.
- **Dead code**: `learned` is only ever assigned `false`, so the "takes a version segment" log arm is unreachable and `UsesVersionSegment = true` is unwritable by the app.
- The 20-second budget now covers both requests, so a slow 404 starves the retry.

Fable verified every individual claim in the original PR body as true — probe gating, URL-in-hand preservation, no-op re-reporting, forget-on-retype. The guarantee that matters, "Perplexity chat works afterwards", is false.

## Where this leaves #742

Conventions are **per-route**, so any one-convention-per-connection model is wrong by construction. Awaiting a decision on the redesign before touching this branch further.

fsnow:
Closing: #742 is out of beta 6 and the design in this branch is wrong regardless — it infers the chat route's convention from the listing route's, and Perplexity uses different conventions for each. Reasoning recorded on #742. Branch kept; reopen if the SDK route-mining approach revives it.

Issue #563, opened 2026-08-06
Dictionaries (DPD and DPPN) not visible on first launch; appear after opening Settings or restarting


Antonio-Costanzo:
Description:

When launching CST Reader 5.0.0-beta.5 for the first time on Windows, the dictionary panel (Ctrl+D) shows only the bundled VRI dictionaries. DPD and DPPN are missing.

Steps:

Launch CST Reader and open a book.

Open dictionary panel (Ctrl+D) → DPD and DPPN are absent.

Go to Settings ▸ Dictionary → DPPN becomes visible immediately.

Close and restart the application → DPD is now visible.

Expected behavior: All available dictionaries should be visible on first launch.

---

Part of epic #109 (third-party dictionary support: DPD + DPPN).

Already fixed on `main` by #536 / PR #539, which landed two days after the beta 5 tag — see the analysis comment below. Kept open to **verify on beta 6**, on Windows, that a clean first run shows both dictionaries without visiting Settings.

fsnow:
Thanks Antonio — this is a known first-run defect that was **already fixed after beta 5 shipped**, so it is not a regression and needs no new work. Recording the detail so the link is unambiguous.

## It is #536, and the fix missed beta 5 by two days

| | |
|---|---|
| beta 5 tag `v5.0.0-beta.5` | `accd7c07` — **2026-07-28** |
| the fix (#536 / PR #539) | `f0a178f9` — **2026-07-30** |

`ReopenableLemmaProvider.cs` does not exist in the beta 5 tree at all. So beta 5 has the bug by construction, and `main` has the fix — it will ship in **beta 6**.

## Why the two dictionaries behaved differently

Antonio's sequence — *"Settings ▸ Dictionary → DPPN becomes visible immediately"* but *"restart → DPD is now visible"* — matches #536's analysis exactly, and the asymmetry is the tell:

- **DPD** was the hard blocker. `ILemmaProvider` bound its availability **in its constructor** and was registered as a DI singleton, so a database that landed after startup left DPD unavailable for the entire session no matter what was on disk. Only a restart could fix it.
- **DPPN** could already flip, because the registry evaluates availability live — but nothing *told the picker to re-query*. Opening Settings happened to rebuild the list as a side effect, which is why that step appeared to work.

[...]

Issue #536, opened 2026-07-30
First run: downloaded dictionaries (DPD, DPPN) don't appear until the app is restarted


fsnow:
Found in beta 5 final testing on a **clean installation**. Sequence: download the Tipiṭaka XML → build the index → download the dictionaries. The dictionaries install successfully, but do not appear in the dictionary panel until the next launch.

This is the **first-run path**, so every new beta 5 user hits it. Not a showstopper — a restart fixes it — but the first impression is "the dictionaries I just waited for aren't there".

## The files really are installed

The install is **in place**, not staged:

```csharp
File.Move(tmp, finalPath, overwrite: true);
```

Staging to `.pending` only happens in the exception branch, when the target can't be replaced because the live db is open (Windows, #394). On a clean install there is no existing db and no open handle, so the asset is live on disk immediately. **The restart is not needed to get the file into place — it is needed because nothing re-reads it.**

## Three layers, two of which are stale

**1. DPD — the provider is opened once at startup.**
```csharp
public bool IsAvailable => _lemma.IsAvailable;   // DpdDictionarySource
```
`ILemmaProvider` is opened at startup (see the note in `App.axaml.cs`: *"Opened once at startup"*). If `dpd-cst-subset.db` was absent then, the provider stays unavailable for the whole session no matter what lands on disk afterwards. **This is the hard blocker for DPD.**

**2. DPPN — probably able to flip, but untested.**
```csharp
public bool IsAvailable => Meta() is not null;   // SqliteDictionarySource
```
`Meta()` calls `SyncStamp()` and re-reads when the file stamp changes, so this one may well go live on its own. Worth confirming rather than assuming.

**3. The picker never rebuilds.** `DictionaryViewModel.RebuildSources()` runs at construction, on a source-preference change, and in `ApplyState` — **nothing calls it when a download completes**. So even a source that correctly flips `IsAvailable` will not appear in the list.

The registry itself is fine for this case: `DictionarySourceFactory.Build` adds `DpdDictionarySource` and the DPPN `SqliteDictionarySource` **unconditionally**, so the entries exist from startup and only their availability is dynamic. (Per the note in `App.axaml.cs`, a *flat-file* directory added mid-session genuinely does need a restart — but that is not this case; the bundled ones are seeded before the registry resolves.)

## Related: the log line says "restart" even when it didn't stage

```csharp
_logger.LogInformation("Installed {Id} (source {Src}, converter {Conv}, ~{Mb} MB). Active on next launch.", ...)
[...]

PR #772, opened 2026-08-21
Cover the first-run dictionary sequence with tests (#563)

fsnow:
## What #563 actually needs

[#563](https://github.com/fsnow/cst/issues/563) is a beta 5 report — a clean Windows install shows neither DPD nor DPPN; DPPN appears after opening Settings, DPD only after a restart. It is #536, fixed by #539 two days after the beta 5 tag, and the issue is held open to **verify on beta 6**.

Verification was hand-only, on Windows. The fix shipped with no tests: every piece was covered on its own, the *join* was not — which is why the defect could only be found by a person doing a clean install.

## What this adds

**`ReopenableLemmaProviderTests`** — the class at the heart of #536 had no coverage at all. Its real contract is that `Reopen` swaps the inner provider **in place**: `LemmaSearchService`, `LemmaReportService` and `DpdDictionarySource` each capture the `ILemmaProvider` reference at construction, so returning a replacement instead would bring the bug straight back, silently, and only on a machine where the asset lands after startup.

**`FirstRunDictionaryVisibilityTests`** — assembles the chain the app wires in `App.axaml.cs` (the registry from `DictionarySourceFactory` over a `ReopenableLemmaProvider`, the picker list from `DictionarySourcePreferenceService`) and drives it with the same two moves the `AssetInstalled` handlers make. The reported asymmetry is pinned as two separate tests so a regression names which half broke:

- the lexicon source goes live on the file alone (it re-stamps on change)
- the DPD source stays dark until the lemma provider is reopened

Plus: an arriving asset is appended without moving the reader's default source, and without undoing a source they had disabled.

**Two installer tests** for the guarantee the whole thing rests on. `UpdateOneAsync` announces `AssetInstalled` only when `!staged`, so if a first install could ever stage, the dictionaries would stay invisible until the next launch — the reported symptom exactly. It cannot: staging lives in the catch below a failed in-place replace, and that catch requires `File.Exists(finalPath)`. With nothing installed there is nothing to lock. That was an argument in the issue thread; it is now an assertion.

## Verified by mutation

Neutering `Reopen()` fails **6 of the 10** new tests. Full suite: **2310 passed, 0 failed**.

## What is still hand-only

The one Windows-only branch these cannot reach: an install that cannot replace a *locked* database stages instead and stays deliberately quiet. A clean first run has nothing to lock — which the installer tests now assert — so the remaining manual check on beta 6 is Antonio's exact sequence: launch clean, let the dictionaries download, open the panel **without** visiting Settings, confirm both are present.

Tests only. No production code changes.

### DIFF
--- a/src/CST.Avalonia.Tests/Services/DpdUpdateServiceTests.cs
+++ b/src/CST.Avalonia.Tests/Services/DpdUpdateServiceTests.cs
@@ -247,6 +247,39 @@ public sealed class DpdUpdateServiceTests : IDisposable
 
     // ---- per-asset version reads ----
 
+    // #563 / #536: a FIRST run cannot stage, structurally. Staging only happens in the catch below a failed
+    // in-place replace, and that catch requires `File.Exists(finalPath)` — so with nothing installed there is
+    // nothing to lock and nothing to fail. That matters because UpdateOneAsync announces AssetInstalled only
+    // when `!staged`: if a first install could ever stage, the dictionaries would stay invisible until the
+    // next launch, which is exactly the reported symptom.
+    [Fact]
+    public void InstallFromGzip_into_an_empty_directory_reports_the_asset_live_not_staged()
+    {
+        var final = Path.Combine(_dir, "first-run", "dpd-cst-subset.db");
+        var gz = Gzip(BuildAssetDbBytes("v0.4.20260531", "3"));
+
+        var staged = DpdUpdateService.InstallFromGzip(gz, Sha(gz), final, DpdUpdateService.ProbeDpdUsable);
+
+        Assert.False(staged);                                    // → AssetInstalled fires
+        Assert.True(File.Exists(final));                         // → and the asset really is live
+        Assert.False(File.Exists(final + ".pending"));
+    }
+
+    // The same guarantee for the lexicon asset, because Antonio saw DPPN and DPD behave differently and the
+    // two travel through separate descriptors.
+    [Fact]
+    public void InstallFromGzip_of_a_first_lexicon_reports_the_asset_live_not_staged()
+    {
+        var final = Path.Combine(_dir, "first-run", "dppn.db");
+        var gz = Gzip(BuildLexiconDbBytes("2025-06", "1"));
+
+        var staged = DpdUpdateService.InstallFromGzip(gz, Sha(gz), final, DpdUpdateService.ProbeLexiconUsable);
+
+        Assert.False(staged);
+        Assert.True(File.Exists(final));
+        Assert.False(File.Exists(final + ".pending"));
+    }
+
     [Fact]
     public void ReadDpdVersion_reads_versions_from_a_real_asset()
     {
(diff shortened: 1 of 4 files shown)


fsnow:
## Fable review — findings and what changed

Two of the five findings were real holes in the **notification layer**, which the first version of this PR simulated rather than exercised. Both would have let the reported symptom return with every test green. Fixed in `2a1eca3`, which is why this is no longer tests-only.

### 1. An ordering race between the reopen and the picker rebuild

`DictionaryViewModel` subscribes to `AssetInstalled` when the layout is built at startup (`App.axaml.cs:219` → `CstDockFactory`); the reopen was bound later, inside the background update check. Multicast handlers run in subscription order, so the picker's handler ran **first** — and it does `Dispatcher.UIThread.Post(...)` and returns.

If the UI thread dequeued that rebuild before `Reopen()` finished opening the database, `DpdDictionarySource.IsAvailable` still read the old inner provider, DPD was omitted, and nothing would rebuild it again. On a clean first run this is usually masked because the dppn install fires a second event moments later — but not when only dpd installs (dppn already current, or its download failed).

The binding now happens right after the service provider is built, before any view model exists, so it is first in the list and swaps synchronously.

### 2. A silent no-op that would revive #536 invisibly

The handler pattern-matched the registered provider and did nothing when it was not reopenable. A DI registration changed back to a bare `SqliteLemmaProvider` — a different file, edited for different reasons, possibly years apart — would have brought #536 straight back with **no error, no log, and no failing test**.

`BindLemmaReopen` is now a testable seam that returns whether it bound and logs an error naming the registered type when it could not. Four new tests cover the binding itself.

### 3. A test that could not fail

`Dispose_releases_every_instance_it_has_opened` called `ClearAllPools()` before `File.Delete`, which released the very handles it claimed to be checking for — it would have passed against a `Dispose` that did nothing, on Windows too. Removed rather than patched; a test that cannot fail is worse than no test.

### Also
- Stopped pinning `DictionarySourceFactory`'s registration order in an assertion that is about *appending*, not ordering.
- Fable confirmed the **cannot-stage** claim is airtight as written, and checked the worry that the subscription sat behind `await xmlUpdateService.CheckForUpdatesAsync()` — that call swallows its own exceptions, so it could not skip the subscription. The move in this commit makes the point moot anyway.

### Verified by mutation
- Restoring the silent-no-op shape → **2** of the new binding tests fail.
- Neutering `Reopen()` → **6** of the others fail.
- Full suite: **2313 passed, 0 failed**.

### Still hand-only on beta 6
Antonio's exact sequence on Windows: launch clean, let the dictionaries download, open the panel **without** visiting Settings, confirm both DPD and DPPN are present.

### Worth filing separately (not in this PR)
Fable noted that a *failed* first install — bad SHA, failed probe, network — is silent for the whole session, because `CheckAndUpdateAsync` runs once per launch with no retry. The reader sees the #563 symptom by a completely different mechanism, and nothing here covers that story.

PR #801, opened 2026-08-21
Try again for a dictionary that failed to download (#773)

### LINKED ISSUE #773: A failed dictionary download is silent for the whole session, with no retry
Found while reviewing #563/#772. Separate mechanism, same visible symptom.

## What happens

`DpdUpdateService.CheckAndUpdateAsync` runs **once per launch**, fired from the startup path. Inside it, a per-asset failure is caught and logged as non-fatal so the other assets still run:

```csharp
catch (Exception ex)
{
    anyFailed = true;
    _logger.LogWarning(ex, "Update of dictionary '{Id}' failed (non-fatal; others continue).", desc.Id);
}
```

That is right as far as it goes — one bad asset should not take down the others. But there is no retry, and nothing re-runs the check later in the session. So a first run whose download fails for any reason — a SHA mismatch, a failed usability probe, a dropped connection, a GitHub hiccup — leaves the reader with **no DPD and no DPPN until they restart the app**.

Which is exactly what #563 reported, arriving by a completely different route. #536 fixed the case where the asset lands and the app fails to notice. This is the case where the asset never lands, and the app never tries again.

## Why it is worth separating from #563

The two are indistinguishable from the reader's chair — the dictionaries are missing and a restart fixes it — but they need different fixes, and the tests added in #772 cover only the first. A regression in this path would look like #563 returning and would send whoever picks it up back through the reopen machinery, which would be fine.

## What the reader currently sees

`StatusChanged` reports `"Could not update some dictionary data."` once, on the welcome page, during startup — while the reader is most likely looking at something else. After that the dictionary panel simply does not list DPD or DPPN, with no indication that anything was attempted or that a restart would help.

## Shape

Undecided, and worth thinking about rather than reaching for the obvious:

- A bounded retry within the session is the obvious move, but the interesting question is *when* — an immediate retry is unlikely to help for the failure modes that actually occur here.
- A failed download is arguably the one case where the dictionary panel should say something, since the picker's silence is currently indistinguishable from "this build has no DPD".
- Neither of these is a first-run-only problem: the same silence applies to a failed *update* of an already-installed asset, where the reader keeps a working older copy and has no reason to suspect anything.

Backend, so filing rather than fixing.

fsnow:
Closes #773.

## The defect

`CheckAndUpdateAsync` runs **once per launch** and catches each asset's failure as non-fatal so the others still run. That is right as far as it goes — one bad asset should not take down the rest. But there is no retry and nothing re-runs the check later, so a first run whose download fails leaves the reader with **no DPD and no DPPN until they restart**.

Which is #563's symptom, reached by a completely different mechanism. #536 fixed the case where the asset lands and the app fails to notice; this is the case where the asset never lands and the app never tries again. Indistinguishable from the reader's chair, and the tests added for #563 cover only the first — a regression here would look like #563 returning and send whoever picks it up back through the reopen machinery, which is sound.

## A failed download is now a state

`AssetFailed` fires and `FailedAssetIds` records it, as a counterpart to `AssetInstalled`.

Without that, a failure is indistinguishable from a build that never offered the dictionary: the picker simply does not list it, which reads as *"this app has no DPD"* rather than *"we could not fetch it"*. Absence of the dictionary was standing in for two entirely different facts.

Recorded **only when the reader has nothing**. A failed *update* leaves a working older asset in place, and reporting that as a missing dictionary would be false — the distinction is `File.Exists(desc.InstallPath)` at the moment of failure. Cleared when a later attempt succeeds.

## The retry is demand-driven, not timed

`RetryMissingAsync`, called when the dictionary panel opens.

Opening the panel is the reader saying they want a dictionary, and that is a better trigger than any interval we could pick: it fires exactly when the absence matters, and it costs nothing otherwise — with both assets present it returns before touching the network, which matters because it runs on every panel open. Fire-and-forget, so the panel opens at once whatever the network is doing, and `AssetInstalled` already brings the picker up to date if the retry succeeds.

An immediate or interval retry would have been the obvious move and a worse one: the failure modes here (offline, unreachable release, dropped connection) are not fixed by trying again a second later, and a timer would spend the reader's bandwidth on a schedule nobody asked for.

## What is deliberately not here

**A visible message.** The state exists so one can be built on it, and the wording belongs to whoever designs the panel. Shipping a half-designed sentence would be worse than shipping the state and saying so.

## Tests

Two new, plus the existing `FirstRunDictionaryVisibilityTests` fake updated for the interface. The interesting one asserts that `RetryMissingAsync` returns **without touching the network** when both assets are present — the property that makes calling it on every panel open acceptable.

Full suite: **2443 passed, 0 failed, 17 skipped.**

### DIFF
--- a/src/CST.Avalonia.Tests/Services/DpdUpdateServiceTests.cs
+++ b/src/CST.Avalonia.Tests/Services/DpdUpdateServiceTests.cs
@@ -1,4 +1,6 @@
 using System;
+using System.Threading.Tasks;
+using CST.Avalonia.Services;
 using System.IO;
 using System.IO.Compression;
 using System.Linq;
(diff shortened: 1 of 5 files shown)


fsnow:
Closing rather than fixing in place. Fable's review found the central mechanism is inert, and I verified both decisive claims myself.

**The retry never fires on panel open.** `DictionaryViewModel` is registered `AddSingleton` (`App.axaml.cs:1492`), so its constructor runs once, during the startup layout build. Every reopen path — `LayoutViewModel.ShowDictionaryPanel`, `DictionaryPanel.axaml.cs` — resolves that same instance. "Recreate on demand" means re-adding the same singleton tool to the layout; the constructor does not re-run.

So `RetryMissingAsync` executes once at startup, milliseconds from the existing check, with `_busy` making one of the pair a no-op. **The user-visible behaviour is unchanged from before the PR.** The exact scenario #773 describes — offline at launch, online later, open the panel — still yields nothing. The comment I wrote describing the trigger describes wiring that does not exist.

**The tests write into the real user data directory.** `DpdUpdateService.DpdSubsetPath` resolves through `AppConstants.DataDirectory` to `~/Library/Application Support/CSTReader`, not a temp dir, and the fixture cleans up only its temp directory. On a machine without real assets that plants an openable but empty `dpd-cst-subset.db` — a ghost dictionary listed in the picker and returning nothing — and a one-entry `dppn.db` stamped `source_version 2025-06`, which matches the shipped DPPN closely enough that the updater could treat the fake as current indefinitely. My own machine escaped only because a `File.Exists` guard saw the real files.

Two more that matter for the redesign:

- **The failure state misses the headline case.** `_failed` is recorded only in the per-asset catch. Offline at launch dies earlier — at the release lookup — so `FailedAssetIds` stays empty and `AssetFailed` never fires, though both dictionaries are missing. The interface comment lists "an unreachable release" as covered. It is not.
- **The ctor retry runs before settings load**, so it sees `EnableAutomaticUpdates = true` regardless of the reader's choice, and can start a ~60 MB download alongside the first-run corpus download — the contention the existing startup sequencing was arranged to avoid.

#773 stays open. The retry is still the right idea; it needs a real activation signal rather than a singleton constructor, a failure state that covers whole-run failures, and an injectable install path so the tests stop writing to a real user's data directory — that last one is a testability defect in `DpdUpdateService` that this PR papered over rather than caused.

PR #837, opened 2026-08-23
Give the book tab the script's font, like every other control (#836)

### LINKED ISSUE #836: Gurmukhi book titles render as tofu in the tab: the tab is the one control with no script font
Reproduced on **Caracara and Kestrel** (macOS x64 and arm64). A book tab whose script is Gurmukhi shows its title as a row of tofu boxes. The book content, the chapter dropdown and the script selector all render Gurmukhi correctly in the same window.

## Cause

The tab title is the only Pāli-bearing control in the window that is **not given a script font**.

Everything else binds one explicitly:

- `BookDisplayView.axaml:94, 225, 297` — `converters:FontHelper.DynamicFontFamily="{Binding CurrentScriptFontFamily}"`
- `OpenBookPanel.axaml:134`, `SearchPanel.axaml:293, 340, 350`, `DictionaryPanel.axaml:71` — the same

The dock tab does not. `Styles/DockStyles.axaml:84` styles `dock|DocumentTabStripItem` for `Background`, `Foreground` and `BorderBrush` and sets no `FontFamily`, so the title renders in the theme's default UI font. `BookDisplayViewModel.CurrentScriptFontFamily` (`:774`) already exists and is already what every other control uses — nothing routes it to the tab.

On macOS that font has no Gurmukhi coverage, and **Avalonia's font fallback does not rescue it** the way native text rendering would, so the glyphs come out as boxes rather than silently substituting Gurmukhi MN.

That the dropdown two inches away renders the same text correctly is the clearest evidence: same window, same script, same string — one control has the binding and the other does not.

## Why Gurmukhi specifically

The default UI font covers most of the scripts this app displays, so the missing binding has been invisible. Gurmukhi is where the coverage gap happens to fall. Worth checking the other twelve rather than assuming this is the only one — the defect is "no script font on the tab", not "Gurmukhi is special", and any script the theme font does not cover will do the same thing.

## Fix

Give `DocumentTabStripItem` the same treatment as every other control — the script font from the document it represents. `CurrentScriptFontFamily` already raises change notifications when the script switches (`:243`, `:2168`), so the tab should follow a script change without further plumbing.

Two things to get right:

- The tab strip also carries non-book documents (Welcome), whose view models have no `CurrentScriptFontFamily`. A style-level binding must degrade to the theme font there rather than erroring.
- It should survive float and unfloat, where the tab is recreated in another window.

## Not yet checked

The same screenshot shows a heading rendered in Devanāgarī (`चुलसीलं`) while the reading script is Gurmukhi. That is content, not chrome, so it is a different question from this issue — but it should not be Devanāgarī either. Filing separately once confirmed rather than folding a guess in here.

fsnow:
Closes #836.

A Gurmukhi book title rendered as tofu in its tab on **both** Caracara and Kestrel, while the same string in the chapter dropdown beside it rendered correctly.

## Cause

The tab was the only control showing Pāli **without a script font**.

| control | script font |
|---|---|
| chapter dropdown, status line (`BookDisplayView.axaml:94, 225, 297`) | ✅ `FontHelper.DynamicFontFamily` |
| book tree (`OpenBookPanel.axaml:134`) | ✅ |
| search, dictionary (`SearchPanel.axaml:293/340/350`, `DictionaryPanel.axaml:71`) | ✅ |
| **document tab** (`DockStyles.axaml:84`) | ❌ Background, Foreground, BorderBrush — no font |

So titles rendered in the theme's UI font, which has no Gurmukhi. Avalonia does not fall back to a family that does, the way native text rendering would, so the glyphs came out as boxes.

**The symptom looks script-specific and isn't.** The theme font happens to cover the other scripts this app displays; any script it did not cover would do exactly the same thing. The maintainer confirmed Gurmukhi renders correctly everywhere else — tree, dropdown, content — which is what isolates it to this one control rather than to the script.

## Fix

```xml
<Style Selector="dock|DocumentTabStripItem" x:DataType="vm:BookDisplayViewModel">
    <Setter Property="FontFamily" Value="{Binding CurrentScriptFontFamily}"/>
</Style>
```

Bound to the **document's own** `CurrentScriptFontFamily`, not the global script setting — a tab shows the script *its* book is displayed in, and per-tab script is deliberate: the toolbar's script is the default for newly opened books, not a statement about the ones already open. The property already raises change notifications on re-scripting, so the tab follows without new plumbing.

`x:DataType` on the `Style` because compiled bindings are on by default and a Style has no other way to know its starting type. `SearchPanel.axaml:340` already binds this same string straight to `FontFamily`, so the string→`FontFamily` conversion is a proven path here.

## Verification

Build clean; suite **2610 passed, 0 failed, 17 skipped**.

**This needs eyes, not tests.** It is a rendered-glyph fact and the project has no headless Avalonia harness. Two things to look at:

1. A Gurmukhi book tab shows its title, and switching that book's script changes the tab with it.
2. **The Welcome tab still looks normal.** It carries a different view model, so the binding finds nothing and the setter should leave `FontFamily` alone — but that is the failure mode worth confirming rather than assuming.

### DIFF
--- a/src/CST.Avalonia/Styles/DockStyles.axaml
+++ b/src/CST.Avalonia/Styles/DockStyles.axaml
@@ -106,6 +106,11 @@
          FontFamily alone, and they keep the theme font, which is what they want anyway. -->
     <Style Selector="dock|DocumentTabStripItem" x:DataType="vm:BookDisplayViewModel">
         <Setter Property="FontFamily" Value="{Binding CurrentScriptFontFamily}"/>
+        <!-- Size travels with the family, because everywhere else it does: every control binding
+             DynamicFontFamily binds DynamicFontSize beside it, and SearchPanel.axaml:17 writes that down
+             as the rule for new chrome. Applying one without the other would make the tab the sole place
+             where the reader's Pāli font is half-honoured — a title in their face at the theme's size. -->
+        <Setter Property="FontSize" Value="{Binding CurrentScriptFontSize}"/>
     </Style>
 
     <!-- Hover (unselected tool tab): stock theme already sets the accent-med background; add the
--- a/src/CST.Avalonia/Styles/DockStyles.axaml
+++ b/src/CST.Avalonia/Styles/DockStyles.axaml
@@ -1,6 +1,7 @@
 <Styles xmlns="https://github.com/avaloniaui"
         xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
-        xmlns:dock="using:Dock.Avalonia.Controls">
+        xmlns:dock="using:Dock.Avalonia.Controls"
+        xmlns:vm="using:CST.Avalonia.ViewModels">
 
     <!-- Position scrollbar below tabs using transform -->
     <Style Selector="dock|DockControl ScrollBar:horizontal">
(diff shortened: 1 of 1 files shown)


fsnow:
## Follow-up: the size travels with the family (`519c966`)

Maintainer noticed the tab took the reader's Pāli *face* but not their *size*, once the glyphs rendered.

That was my omission, and the codebase is unambiguous about it: every control binding `DynamicFontFamily` binds `DynamicFontSize` beside it — `OpenBookPanel:135`, `BookDisplayView:95/226/298`, `SearchPanel:294/341/351`, `DictionaryPanel:72` — and `SearchPanel.axaml:17` states it as the rule for new chrome. Applying one without the other left the tab as the sole place where the setting is half-honoured.

```xml
<Setter Property="FontSize" Value="{Binding CurrentScriptFontSize}"/>
```

**One thing to look at that the family alone did not raise:** the tab strip has to accommodate whatever size is set. At a large Pāli size the strip will grow, and the question is whether it grows gracefully or clips. Worth trying a deliberately large size in Settings ▸ Pāli Fonts before merging — if it clips, the answer is probably a cap on this setter rather than dropping it, since a tab that ignores the size is the state just fixed.

Issue #278, opened 2026-07-11
AI integration: MCP via app --mcp-bridge (stdio relay to the running app's /mcp) — stable client config with an ephemeral API


fsnow:
Path B from the design discussion. Give an MCP chat client (Claude Desktop) a **static, value-free config** while keeping the local API **ephemeral + per-session token** (the secure fresh-`local-api.json`-read model that code agents already use). No long-lived secret, no fixed port, no separate binary, no Node.

## The idea
A static MCP client config (via `mcp-remote`) must bake in a fixed URL + token, which forces the whole API onto a long-lived token + fixed port (what #275/#276 did). Instead, front the MCP client with a bridge that reads `local-api.json` at spawn time — and make that bridge **CST Reader's own already-signed binary**, launched with a flag:

```json
{ "mcpServers": { "cst-reader": {
    "command": "/Applications/CST Reader.app/Contents/MacOS/CST",
    "args": ["--mcp-bridge"]
} } }
```

> **Bundle path (verified on the signed build):** the macOS bundle's `Contents/MacOS/` holds `CST` (the `CFBundleExecutable` launch wrapper, which `exec`s `./CST.Avalonia "$@"` — so it forwards `--mcp-bridge`) and `CST.Avalonia` (the .NET host). There is **no** `CST Reader` binary. Hand configs should point at `.../MacOS/CST`. (Note for the Phase 4 Copy-config helper: inside the .NET process `Environment.ProcessPath` resolves to `.../MacOS/CST.Avalonia`, not the `CST` wrapper — both are valid launch targets, so emit whichever the helper can name reliably.)

On `--mcp-bridge`, the app skips the GUI/single-instance path, reads `local-api.json` (the running app's current port + token), and runs a **stdio ↔ `/mcp` relay** to the running app. The client config carries no port/token, so it survives every restart; the API can stay ephemeral.

## Scope
- `--mcp-bridge` entry: detect the flag BEFORE Avalonia/single-instance init; run the relay; exit when stdin closes.
- Read `local-api.json`; if the app is not running / API is off, return a clear MCP tool error (do not hang).
- The relay: connect to `/mcp` (Streamable HTTP + bearer) as an MCP client and re-expose its tools/resources on a stdio MCP server, forwarding calls (a proxying MCP server).
- "Copy MCP configuration" (Settings) emits THIS config, not the `mcp-remote` snippet.
- Gated by the `McpEnabled` toggle (serve `/mcp` only when on).
[...]

PR #858, opened 2026-08-24
A sample prompt for agents that speak HTTP rather than MCP (#857)

### LINKED ISSUE #857: Local API settings: a "Sample prompt" expander to orient a coding agent
**[fsnow]** The local-API settings should carry a **"Sample prompt"** expander — the minimum a reader can paste into a coding agent to get it oriented to the API, including the path to the local file holding the key.

## The gap

The MCP path is well served: a Copy-configuration button and a Show-configuration expander, ready to paste into Claude Desktop.

An agent that speaks plain HTTP instead — Claude Code, or anything driving `curl` — has nothing. The reader has to know that `local-api.json` exists, where it lives on their platform, that it holds `port` and `token`, that the token goes in an `Authorization: Bearer` header, and that `/llms.txt` describes the rest. All of that is written down somewhere; none of it is in the app.

## What it should say, and what it should not

**It must name the handshake file by its real path**, computed rather than written down, so a Windows reader is not shown a macOS path.

**It must point at `/llms.txt` rather than list the endpoints.** The server already serves a thin index and `/llms-full.txt` for detail. A copy of the route list living in a settings string is one that goes stale the first time a route is added — silently, in the place nobody thinks to check.

**It must tell the agent to read the file each time rather than remember it.** Port and token change on every start (#278). An agent that caches them works until the app restarts and then fails in a way that looks like the API is broken.

That is four facts and a pointer. Anything more duplicates `/llms.txt`.

## Placement

Beside the MCP configuration block, as a second expander with its own Copy button — same shape, same confirm flash. The two are alternatives: MCP for a client that speaks it, the prompt for an agent that does not.

fsnow:
Closes #857.

The MCP path already had a Copy-configuration button and a Show-configuration expander. An agent that speaks plain HTTP — Claude Code, or anything driving `curl` — had nothing.

## What the prompt carries

Four facts and a pointer:

1. **The handshake file, by its real path** — computed at read time, so a Windows reader is not shown a macOS path.
2. **What is in it** — `port` and `token`.
3. **How the token is used** — `Authorization: Bearer <token>` against `http://127.0.0.1:<port>`.
4. **Read it each time, do not remember it** — port and token change on every start (#278). An agent that caches them works until the app restarts, then fails in a way that looks like the API is broken.
5. **Then fetch `/llms.txt`.**

## What it deliberately does not carry

**The endpoint list.** The server already serves `/llms.txt` (thin index) and `/llms-full.txt` (detail) for exactly this purpose. A copy of the route list living in a settings string is one that goes stale the first time a route is added — silently, in the place nobody thinks to check.

Anything beyond the five points above duplicates `/llms.txt`, which is the thing designed to be authoritative.

## Placement

A second expander beside the MCP configuration, with its own Copy button and confirm flash, mirroring `OnCopyMcpConfig`. The copy handler is in code-behind for the same stated reason as the existing one: the clipboard needs a `TopLevel` and the view model has none.

The two are alternatives — MCP for a client that speaks it, the prompt for an agent that does not.

## Verification

Build clean; suite **2611 passed, 0 failed, 17 skipped**.

**No test.** The property is a formatted string with no branching worth pinning, and the expander and its copy button are rendered facts this project has no headless harness for.

Worth checking by eye: the path shown is the real one for the platform, the Copy button flashes its confirm, and the prompt actually works — paste it into a coding agent with CST Reader running and see whether it gets to `/llms.txt` unaided. That last one is the only test that matters.

### DIFF
--- a/src/CST.Avalonia/ViewModels/SettingsViewModel.cs
+++ b/src/CST.Avalonia/ViewModels/SettingsViewModel.cs
@@ -1548,6 +1548,55 @@ public class AiSettingsViewModel : ViewModelBase, IDisposable
         public string McpClientConfigJson => CST.Avalonia.Services.LocalApi.McpClientConfig.ClaudeDesktop(
             System.Environment.ProcessPath ?? "CST Reader");
 
+        /// <summary>
+        /// A prompt a reader can paste into a coding agent to get it talking to the local API.
+        ///
+        /// <para><b>It points at <c>/llms.txt</c> rather than listing the endpoints.</b> The server already
+        /// serves a thin index and a full one for exactly this purpose, and a copy of the route list living
+        /// in a settings string is a copy that goes stale the first time a route is added — silently, and in
+        /// the one place nobody thinks to check.</para>
+        ///
+        /// <para>The handshake path is computed, not written down, so it names the real file on whichever
+        /// platform is asking rather than a macOS path shown to a Windows reader.</para>
+        ///
+        /// <para><b>It tells the agent to read the file rather than remember it.</b> Port and token change on
+        /// every start (#278), so an agent that caches them works until the app restarts and then fails in a
+        /// way that looks like the API is broken.</para>
+        /// </summary>
+        public string LocalApiSamplePrompt
+        {
+            get
+            {
+                var path = System.IO.Path.Combine(
+                    System.Environment.GetFolderPath(System.Environment.SpecialFolder.ApplicationData),
+                    AppConstants.AppDataDirectoryName,
+                    Services.LocalApi.LocalApiInfo.FileName);
+
+                return string.Join(System.Environment.NewLine, new[]
+                {
+                    "CST Reader is running on this machine and exposes a local HTTP API over the",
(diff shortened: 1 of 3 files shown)


fsnow:
## Follow-up: the prompt now names no endpoint at all

The first draft told the agent to fetch `/llms.txt`. `local-api.json` already carries a `docs` field for exactly that purpose — `LocalApiInfo.Docs`, commented *"so a client needn't guess"*.

So the prompt was duplicating a value the app publishes, and would have gone quietly wrong the day that value changed. That is the same argument used above for not listing the routes, applied one level further up: **name the file, let the file name the rest.**

It also brings the wording into line with the cold-agent prompt in `docs/testing/ai-prompts/navigate-show-me.md`:

> It contains a port, a bearer token, and the path to the API's own documentation. Read the documentation before you do anything else, and follow it.

That prompt exists to be re-run against new models, as a read on how well this surface teaches an agent that has never met it. The prompt shipped to readers and the prompt under test should not drift apart — otherwise the tested one passes and the shipped one is something else.

Issue #805, opened 2026-08-22
The resolver sends the active model to the wrong connection, and never checks the model is on it


fsnow:
Found in an adversarial review of surface B, then sharpened by a second review pass. Both halves confirmed against `76d5405`.

## Half 1 — the resolver guesses which connection to use

`ChatProviderResolver.Resolve` (src/CST.Avalonia/Services/Ai/ChatProviderResolver.cs:178-181):

```csharp
var connection = chat.Connections.FirstOrDefault(
    c => string.Equals(c.Id, chat.ActiveConnectionId, StringComparison.Ordinal))
    ?? chat.Connections.FirstOrDefault();
```

When `ActiveConnectionId` matches nothing, the request goes to whichever connection happens to be first, carrying `ActiveModelId` — a model that belongs to a different endpoint. The result is a 401 or 404 naming neither cause, which is the #678 shape.

**The service already disagrees with it.** `AiConnectionService.Active` fails closed — `Find(Chat.ActiveConnectionId) is { } record ? ToRuntime(record) : null` (AiConnectionService.cs:210). So when the fallback fires, the model-picker chip reads "No model" while the resolver sends the request anyway. Two surfaces reading one fact and contradicting each other.

**Reachability.** Every in-app writer keeps the pair consistent: `Add`/`AddFromPreset` do `ActiveConnectionId ??= record.Id` (:229, :267), `Remove` resets both (:386-390), `SetActive` writes them together and refuses a model not on the connection (:406-407), `FollowTheChoice` follows enable/disable (:550-559), and `Apply` never changes an id — `Update` declares it immutable (:80). But:

- **A hand-edited `settings.json` is in contract.** The resolver says so itself (ChatProviderResolver.cs:217-218, :344). And since the id is immutable through the UI, hand-editing is the *only* way to rename a connection — so a reader who renames one creates the stale pointer with certainty.
- **`SettingsValidator.Sanitize` removes connections on load** (src/CST.Avalonia/Models/SettingsValidator.cs:140-143) and never validates the two active pointers while repairing everything around them. It is the seam already built for this repair, left unfinished.

Note the fallback is not *always* wrong: with a single connection configured, it picks the only endpoint, which is the right one. The argument for removing it is that failing closed with an accurate message beats self-healing in a way the picker visibly contradicts — not that it never helps.

## Half 2 — nothing checks the model belongs to the connection

[...]

PR #872, opened 2026-08-28
Resolve a lemma the reader capitalised (#868)

### LINKED ISSUE #868: Lemma lookup is case- and normalization-sensitive: `Dhammā` resolves nothing, `dhammā` resolves 17 ways
Found in the 2026-08 Fable code review (`docs/development/CODE_REVIEW_2026-08_FINDINGS.md`, area **R13-2**). Verified against the installed asset.

`LemmaSearchService.ResolveWord` / `Deconstruct` pass the word **verbatim** into `WHERE fl.form = $f` (BINARY collation) when `Script.Latin` — which is the REST and MCP **default**. No lowercasing, no NFC composition, no joiner strip.

```csharp
// src/CST.Avalonia/Services/LemmaSearchService.cs:38-53
```

The asset stores forms lowercase NFC. Verified directly:

```
$ sqlite3 "file:…/dpd-cst-subset/dpd-cst-subset.db?mode=ro" \
    "select count(*) from form_lemma where form='dhammā';"   -- 17
    "select count(*) from form_lemma where form='Dhammā';"   -- 0
```

So `GET /v1/lemma/Dhammā` and MCP `lemma_lookup("Dhammā")` return a confident **"No lemma resolves the form"** for a word the shipped data resolves seventeen ways. NFD-decomposed input misses identically (zero forms in the asset contain combining U+0304).

## Why this is worth fixing first

- **It is reachable by anyone, right now.** Sentence-initial Pāli is capitalised. Any agent echoing a proper-noun casing hits it. There is no error — the answer is a clean, confident negative.
- **Every sibling path already normalizes.** `DpdDictionarySource.cs:47` does StripJoiners → `ToLowerInvariant` → `Normalize(FormC)`; `AiContextBundler` lowercases; `SelectionPipeline` NFC-composes. This one branch is the outlier.
- **The interface doc promises the normalization that the Latin branch never performs** ("The word is normalized to IAST for the DPD lookup").

## Fix

Apply the `DpdDictionarySource` normalization to the input in both methods:

```csharp
form = StripJoiners(form).ToLowerInvariant().Normalize(NormalizationForm.FormC);
```

## Related

This is the third instance of the same class found in this review — a comparison that knows only one case or one normalization form, silently returning "not found":

- `AnswerScorer.FindUnmarkedPali` is blind to NFD Pāli (R2-2)
- `ChatProviderResolver`'s Ordinal id match vs the service's OrdinalIgnoreCase (#805)

Worth a sweep for others rather than three separate point fixes.

fsnow:
Fixes #868.

## The bug

`LemmaSearchService.ResolveWord`/`Deconstruct` passed the word **verbatim** into a byte-exact `WHERE fl.form = $f` (BINARY collation) when `sourceScript == Script.Latin` — which is the REST and MCP **default**. No lowercasing, no NFC, no joiner strip.

Verified against the installed asset:

```
select count(*) from form_lemma where form='dhammā';   -- 17
select count(*) from form_lemma where form='Dhammā';   --  0
```

So `GET /v1/lemma/Dhammā` and MCP `lemma_lookup("Dhammā")` returned a confident **"No lemma resolves the form"** for a word the shipped data resolves seventeen ways. Sentence-initial Pāli does this; so does any agent echoing a capitalised form. NFD input missed identically.

There was no error — the failure is indistinguishable from a word DPD genuinely does not cover.

## The fix

Both branches now normalize through one `ToLookupKey` method: strip joiners → lower-case → NFC. This mirrors what `DpdDictionarySource` and `DictionaryService` have always done; this one branch was the outlier, and the method's own doc comment already claimed the normalization it never performed.

Two deliberate choices:

- **One method for both branches.** They used to differ, and fixing only the Latin path would have left them free to differ again — this is the review's most repeated finding shape (a guard that exists and a sibling that did not adopt it).
- **The non-Latin branch re-normalizes after conversion**, so the postcondition holds whatever a converter emits, rather than resting on the current converters happening to produce composed lower-case output.

## Tests

Five theory cases (as-stored, capital, all-caps, NFD, zero-width non-joiner) plus a non-Latin round-trip and a negative control.

**Mutation-checked:** with the fix reverted, four of the five theory cases fail; the composed lower-case case passes either way, which is what a control should do. The negative control confirms the change widens what resolves rather than making everything resolve.

Full suite: **2599 passed, 0 failed, 17 skipped**.

## Provenance

Found by the 2026-08 Fable code review, area **R13-2**. The review also found `AnswerScorer.FindUnmarkedPali` blind to NFD (R2-2) and the Ordinal/OrdinalIgnoreCase split in #805 — same class, worth a sweep rather than three point fixes.

### DIFF
--- a/src/CST.Avalonia.Tests/Services/LemmaSearchServiceTests.cs
+++ b/src/CST.Avalonia.Tests/Services/LemmaSearchServiceTests.cs
@@ -114,6 +114,47 @@ public sealed class LemmaSearchServiceTests : IDisposable
         Assert.Null(svc.ResolveWord("Nota\u00F1\u00F1word", Script.Latin));
     }
 
+    // The regression the first draft of #868 introduced, caught in review. Script.Ipe is this interface's
+    // DECLARED DEFAULT, and IPE encodes Pāli letters in the Latin-1 *uppercase* block from U+00C0 — so
+    // case-folding before conversion shifts every letter onto a different IPE letter or onto nothing
+    // ("dhammā" read out as "übhmmm"). No production caller reaches the default today (REST rejects Ipe,
+    // MCP coerces it to Latin), which is exactly why the suite stayed green while the default path was
+    // broken. Derived through the converter rather than hard-coded, so this cannot drift from IPE itself.
+    [Fact]
+    public void ResolveWord_resolves_IPE_input_on_the_interfaces_default_script()
+    {
+        var svc = NewService(new FakeSearchService());
+        var ipe = ScriptConverter.Convert("pa\u00F1\u00F1\u0101ya", Script.Latin, Script.Ipe);
+
+        var r = svc.ResolveWord(ipe); // no sourceScript: the Script.Ipe default
+
+        Assert.NotNull(r);
+        Assert.True(r!.Candidates.Count > 1);
+    }
+
+    // Latn2Ipe folds ṁ (U+1E41) and ṃ (U+1E43) together as niggahita, so every other Latin-input path
+    // canonicalises them. Before the Latin branch round-tripped through IPE, `dhammaṁ` resolved in the
+    // dictionary panel and reported "no lemma resolves this form" here — the same silent-negative class
+    // as #868 itself. The fixture stores `paññaṃ` with U+1E43. (fable review)
+    [Theory]
+    [InlineData("pa\u00F1\u00F1a\u1E43")] // ṃ, as stored
+    [InlineData("pa\u00F1\u00F1a\u1E41")] // ṁ, the other spelling
+    public void ResolveWord_folds_both_spellings_of_niggahita(string word)
+    {
+        var svc = NewService(new FakeSearchService());
+
+        Assert.NotNull(svc.ResolveWord(word, Script.Latin));
+    }
+
+    // Surrounding whitespace survives IsNullOrWhiteSpace and used to miss the byte-exact lookup.
+    [Fact]
+    public void ResolveWord_ignores_surrounding_whitespace()
+    {
+        var svc = NewService(new FakeSearchService());
+
+        Assert.NotNull(svc.ResolveWord("  pa\u00F1\u00F1\u0101ya\t", Script.Latin));
+    }
+
     [Fact]
     public async Task ExpandAndSearch_builds_regex_alternation_and_maps_counts()
     {
--- a/src/CST.Avalonia.Tests/Services/LemmaSearchServiceTests.cs
+++ b/src/CST.Avalonia.Tests/Services/LemmaSearchServiceTests.cs
@@ -68,6 +68,52 @@ public sealed class LemmaSearchServiceTests : IDisposable
         Assert.True(r!.Candidates.Count > 1); // homograph
     }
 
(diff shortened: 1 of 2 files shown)


fsnow:
## Adversarial review (Fable) — one merge-blocking regression, now fixed

The review found that the first commit **fixed the common path and broke the documented default one** — the same "confident wrong negative" shape this PR set out to eliminate.

### Finding 1 (HIGH, regression) — fixed in 1e6123e

`Script.Ipe` is `ILemmaSearchService`'s declared **default** for both methods, and IPE encodes Pāli letters in the Latin-1 **uppercase** block from U+00C0 (`Ipe2Latn.cs:22-77`). Case-folding before conversion shifted every letter 0x20 onto a *different IPE letter* or onto nothing — IPE `dhammā` read out as `übhmmm`, verified empirically against CST.Core and reproduced at the service level.

No production caller reaches the default today (REST's `ParseScript` rejects Ipe, MCP's `Safe()` coerces it to Latin), which is precisely why 2599 tests stayed green over a broken default.

The reasoning that produced it was "mirror `DpdDictionarySource`" — but that sibling folds **user text on its way into IPE** and never folds IPE itself. The resemblance was misleading. Case-folding now happens **after** conversion, when the string is certainly Latin.

### Finding 2 (MEDIUM, incomplete fix) — fixed

`Latn2Ipe` folds both spellings of niggahita — `ṁ` (U+1E41) and `ṃ` (U+1E43) — so every other Latin-input path canonicalises them. The Latin branch returned without an IPE round-trip, so `dhammaṁ` resolved in the dictionary panel while reporting "no lemma resolves this form" here. Same silent-negative class as #868 itself. The Latin branch now round-trips through IPE.

### Finding 3 (LOW, pre-existing) — fixed

Leading/trailing whitespace survives `IsNullOrWhiteSpace` and then missed the byte-exact lookup. Trimmed.

### Verified sound by the review

- **The IAST inventory round-trips** under `ToLowerInvariant` + NFC — ā ī ū ṁ ṃ ṅ ñ ṭ ḍ ṇ ḷ ṛ ś ṣ all fold to the right codepoints with no collisions (`ṁ`/`ṃ` stay distinct as characters); `Normalize(FormC)` is identity on stored DPD forms.
- **The non-Latin path is safe.** The Indic scripts are caseless, and every converter already maps ZWJ/ZWNJ to `""` — so stripping joiners pre-conversion removes nothing transliteration-bearing. Checked explicitly for Sinhala, whose `Deva2Sinh` output carries ZWNJ: it and Devanagari both produce the key `paññāya`. This was the risk I flagged as sharpest going in; it is not real.
- **No second-order effects.** The only consumer of `WordDeconstruction.Word` is `LemmaApi.ToDeconstruct`, which ignores it and echoes the caller's original word. API echoes are unchanged.
- **The original mutation check reproduced independently** — 4 of 5 theory cases fail with the fix reverted, and the assertions genuinely discriminate.

### Now

3 new tests (IPE-on-the-default, both niggahita spellings, whitespace), each mutation-checked: restoring the pre-conversion fold fails the IPE test **and nothing else**; removing the round-trip fails the niggahita case **and nothing else**.

Full suite: **2603 passed, 0 failed, 17 skipped**.

Issue #787, opened 2026-08-21
Nothing tests that this version can load what the previous version wrote


fsnow:
The generalisable half of the #784 incident, filed separately because it is not about settings.

## The gap

Every persisted model in this app is round-trip tested: serialize the current shape, deserialize it, assert it survives. Those tests pass the whole time a type change is breaking real installations, because both halves use the *new* shape.

The test that fails is a different one: **does this version load what the previous version wrote?**

Nobody writes it, and the reason is structural rather than careless — after a type change the old shape no longer exists in the code to write a test against. It has to be captured as a literal: a JSON fixture of what the previous version actually produced, checked in, and asserted to still load.

That is the only kind of test that can catch a persisted-type change, and there is not one in the tree.

## What it would have caught

`AiConnectionRecord.Headers` went from `Dictionary<string, string>` to `List<AiHeaderRecord>`. Every round-trip test passed. A settings file written a day earlier failed to parse:

```
JsonException: The JSON value could not be converted to List<AiHeaderRecord>.
Path: $.Ai.Chat.Connections[0].Headers
```

Writing the test afterwards immediately found a second hole in the same place — `"Headers": null` bypassed the converter entirely, because `System.Text.Json` does not consult a converter for null unless it opts in via `HandleNull`.

## Scope

Both persisted stores, not just settings:

- **`Settings`** (`settings.json`) — no such test.
- **`ApplicationState`** (`app-state.json`) — no such test either. It looks protected because it recovers from timestamped backups (`ApplicationStateService.TryLoadFromBackupAsync`), but **the backups are the same shape as the primary file**. A type change breaks them too, and the recovery walks a directory of files it also cannot read. Backup recovery and this test are complements, not alternatives.

## Shape

A small corpus of checked-in fixtures — one per released version, or one per shape change — asserted to deserialize into something usable. Not asserted to be *identical* to what a current version would write: the point is that the load path copes, including with fields that no longer exist and fields that did not exist yet.

Cheap to add, and each fixture is written once and never edited. The discipline is remembering to capture one *before* changing a persisted type, which is the part worth writing down somewhere an agent will read it.

Related: #785 (the blast radius when a load does fail), #319.

Issue #728, opened 2026-08-19
A model removed by its provider is still shown, described, and pickable


fsnow:
Part of #689 / #674.

> **Provenance.** Agent-authored. **[fsnow]** marks the maintainer's words; **[suggestion]** is mine and advisory; **[observed]** is read from the code, with the file named.

## The question that prompted it

**[fsnow]** *"Do we have any handling of the removal of a model by a provider?"*

**[observed]** No. Verified across the three places it would surface:

| | Today |
|---|---|
| **Models tab** (`AiModelsViewModel.Rebind`) | merges the fetched catalogue with the connection's stored `Models[]`, but never compares them. A stored model the provider no longer lists sits there looking normal — enabled, pickable, indistinguishable. |
| **Resolver** (`ChatProviderResolver`) | takes `ActiveModelId` verbatim, with no validation. Deliberate: an arbitrary endpoint may publish no listing at all, so a picker can never be a gate. A removed model resolves fine and fails at the wire. |
| **Provider** | returns 404/400, surfaced as `AiErrorKind.Provider` with whatever sentence the endpoint wrote. The reader gets *something*, but nothing connects it to "the model in your picker no longer exists." |

It was also explicitly deferred once. #723 (removing the "Typed" badge) recorded: *"The repair would be to badge a stored model that the provider's listing does not carry — a stale or renamed id, which a listing genuinely can tell us and which would explain a 404 at send time. But nobody has hit that."* That reasoning was right at the time; the question has now been asked.

## Why it matters more since #726

#726 began caching the provider's published facts on `AiModelRecord` — context length, reasoning, inputs — so the per-turn picker can show them without a fetch.

So a removed model now keeps its hover card. **The app will confidently describe a model that no longer exists**, with a context window and a reasoning flag, in the same shape it describes real ones. That is a worse failure than the silence it replaced, and it is the direct cost of a decision taken for good reasons.

Not hypothetical: OpenRouter retires and renames models routinely, and free models especially come and go.

## Suggested shape

**[suggestion]** Small, and mostly about what NOT to do:

- When a catalogue fetch **succeeds**, compare the listing against the connection's stored `Models[]`. Mark any stored model the listing does not carry.
- **A fetch that fails must mark nothing.** An offline moment, a dead local runner, or a 401 would otherwise flag every model as gone — turning a transient problem into a screen full of false alarms. Same reasoning as `Reachability.Configured` (#673): absence of evidence is not evidence.
[...]

Issue #319, opened 2026-07-12
[MED] A null `ai`/`localApi` section in settings.json permanently bricks the Settings window


fsnow:
`SettingsValidator.Sanitize` (`SettingsValidator.cs:61-122`) repairs `DeveloperSettings`/`XmlUpdateSettings`/`FontSettings` with `??=` but **not** `settings.Ai` or `settings.Ai.LocalApi`.

**Scenario:** a hand-edited/corrupted `settings.json` with `"ai": null` (or `"ai": {"localApi": null}`) deserializes fine (the explicit null overwrites the `= new()` initializer). `AiSettingsViewModel`'s ctor then NREs at `SettingsViewModel.cs:816-821` (`ai.LocalApi.Enabled`); `ShowSettingsWindow`'s catch (`App.axaml.cs:1424-1427`) swallows it → the **Settings window silently never opens again**, and since the only repair path is the Settings UI, the user is stuck until they hand-edit the file. (`StartLocalApiIfEnabledAsync` also NREs, caught+logged.)

**Fix:** add `settings.Ai ??= new AiSettings(); settings.Ai.LocalApi ??= new LocalApiSettings();` to `Sanitize` (matches the existing pattern). (A7-2)

From the Fable-led code review (area A7); full detail in docs/development/AI_INTEGRATION_REVIEW_FINDINGS.md. Part of epic #186.

PR #873, opened 2026-08-28
Refuse a repeated model id, and open over one that is already there (#870)

### LINKED ISSUE #870: Duplicate model ids crash the Settings window, and can lock the reader out of it permanently
Found in the 2026-08 Fable code review (`docs/development/CODE_REVIEW_2026-08_FINDINGS.md`, area **R5-1**, composing with **R4-3**).

## What happens

Nothing refuses two model rows with the same id. `AiModelsViewModel.ApplyFilter` then throws:

```csharp
// src/CST.Avalonia/ViewModels/AiModelsViewModel.cs:393
var stored = _connection.Models.ToDictionary(m => m.Id, StringComparer.Ordinal);
```

`ToDictionary` throws `ArgumentException` on a duplicate key, and this runs inside the **synchronous** `ConnectionsChanged` chain — so Save crashes on the spot.

## The tail is what makes this bad

If the already-armed debounce write commits the duplicate to `settings.json` before the crash, **every subsequent Settings open crashes** in the `AiSettingsViewModel` constructor. The reader cannot reach the UI that would let them undo it. Recovery requires hand-editing `settings.json`.

## Two ways in

1. **Via the editor** — add a model row, paste the same id twice, Save. Nothing dedupes: `TypedModels` doesn't, and `AiConnectionService.Apply` copies the draft verbatim (`AiConnectionEditorViewModel.cs:745-763`).
2. **Without the editor (R4-3)** — a provider listing whose id carries surrounding whitespace. `AiConnectionService.EnableModel` (`:574-584`) looks up by the **untrimmed** id but stores `Id = modelId.Trim()`, so every toggle misses the row the previous toggle created and appends another duplicate. Nothing trims a listing id on the way in (`AiModelCatalog.Parse:563` only requires `Length > 0`). A sloppy gateway or local runner is enough.

## Fix

Two independent guards, both worth having:

- **Refuse duplicate model ids at the editor/service seam**, the way headers already are — `DuplicateHeader` exists; models got nothing, in the same file.
- **Make `ApplyFilter` tolerant** (`GroupBy`/`First`), so a settings file that already contains duplicates opens rather than locking the reader out.

Trim the model id once at entry in `EnableModel`, before the lookup, to close the second route.

## Note

The asymmetry is the tell: headers in this same editor are protected against duplicates and models are not. This review found eight instances of that shape — a guard that exists and a sibling that did not adopt it — summarised in the review document.

fsnow:
Fixes #870.

Two model rows with one id had nothing to stop them. The sheet let a reader paste an id twice, `Add`/`Update` copied the draft verbatim into the record, and `AiModelsViewModel.ApplyFilter` then threw `ArgumentException` keying its stored list with `ToDictionary`. That throw happens inside the `AiSettingsViewModel` constructor — so once the debounced save had committed the pair, **every subsequent Settings open crashed**, and the only screen that could have removed the duplicate was the one that would not open.

### Three parts

The guard alone does nothing for a reader whose `settings.json` already holds a pair, which is why the tolerance is not redundant with it.

| | |
|---|---|
| `AiConnectionService.DuplicateModel` | Refuses a repeat at `Add` and `Update`, the same shape as the existing `DuplicateHeader`. Trimmed + ordinal, matching what the record stores and what a toggle resolves. Refused rather than silently collapsed: the reader typed both rows, and a save that quietly dropped one would look like a lost edit. |
| `AiModelsViewModel.ApplyFilter` | First-wins over the stored list instead of keying it — the way back in for a file that already carries duplicates. First-wins is also the row a toggle acts on, since `EnableModel`/`SetModelEnabled` both resolve an id by `FirstOrDefault`. |
| `AiConnectionService.EnableModel` | Trims the model id **once, before the lookup**. It looked up untrimmed and stored trimmed, so a provider listing id carrying whitespace never matched the previous toggle's record and appended another one on every pass — the same duplicate, by a route no editor-side guard can see (R4-3). |

### Tests

Five added, each mutation-checked (revert the corresponding fix, confirm only that test fails):

- `Two_models_with_one_id_are_refused_on_add` / `_on_update` / `A_padded_repeat_of_a_model_id_is_refused`
- `A_padded_listing_id_does_not_append_a_second_model`
- `A_stored_list_that_already_holds_a_duplicate_id_still_opens` — seeds the duplicate straight into `Settings`, since the service now refuses to create one

Full suite: **2626 passed, 0 failed, 17 skipped**.

Found by the August Fable review (R5-1, with R4-3 as the second route).

### DIFF
--- a/src/CST.Avalonia.Tests/Models/StateValidationTests.cs
+++ b/src/CST.Avalonia.Tests/Models/StateValidationTests.cs
@@ -383,6 +383,37 @@ public class StateValidationTests
     }
 
     [Fact]
+    public void Settings_Sanitize_trims_and_collapses_repeated_model_ids()
+    {
+        // #870: two records with one id used to throw ArgumentException out of the models tab — inside the
+        // Settings window's construction, so the window would not reopen. Repaired at load rather than
+        // tolerated per-screen: the per-turn picker has no dedupe either, and with BOTH records enabled (the
+        // shape both real routes produce) it listed the model twice, while a toggle reached only the first —
+        // so switching it off left the twin on and answering.
+        var s = new Settings();
+        s.Ai.Chat.Connections.Add(new AiConnectionRecord
+        {
+            Id = "mine",
+            Models =
+            {
+                new AiModelRecord { Id = "gpt-4o", DisplayName = "GPT-4o", Enabled = true },
+                new AiModelRecord { Id = " gpt-4o ", DisplayName = "GPT-4o again", Enabled = true },
+                new AiModelRecord { Id = "claude-opus-5", DisplayName = "Opus", Enabled = true },
+            },
+        });
+
+        var fixes = SettingsValidator.Sanitize(s);
+
+        var models = s.Ai.Chat.Connections.Single().Models;
+        Assert.Equal(new[] { "gpt-4o", "claude-opus-5" }, models.Select(m => m.Id));
+        Assert.Equal("GPT-4o", models[0].DisplayName);           // first wins, the row a toggle acts on
+        Assert.Contains(fixes, f => f.Contains("repeated model id"));
+
+        // Idempotent, like every other repair here.
+        Assert.DoesNotContain(SettingsValidator.Sanitize(s), f => f.Contains("repeated model id"));
+    }
+
+    [Fact]
     public void Settings_Sanitize_ClampsNonPositiveFontSizes()
     {
         var s = new Settings();
--- a/src/CST.Avalonia.Tests/Services/Ai/AiConnectionServiceTests.cs
+++ b/src/CST.Avalonia.Tests/Services/Ai/AiConnectionServiceTests.cs
@@ -44,6 +44,83 @@ public class AiConnectionServiceTests
         Assert.Contains("already", second.Problem);
     }
 
+    // ---- duplicate model ids (#870) --------------------------------------------------------------------
+
+    private static AiConnectionDraft DraftWith(params AiModelEntry[] models) =>
+        new("My box", ChatProviderKind.OpenAiCompatible, "http://localhost:8000/v1",
+            models, Array.Empty<AiHeader>(), new Dictionary<string, string>());
+
+    /// <summary>
+    /// Two rows with one id are refused on the way in, the way two headers with one name already were.
+    ///
+    /// <para>Nothing used to stop them: the sheet let a reader paste an id twice, the draft was copied
+    /// verbatim into the record, and the models tab then threw <c>ArgumentException</c> keying its stored
+    /// list — inside the Settings window's own construction, so the window would not open again. (#870)</para>
+    /// </summary>
+    [Fact]
+    public void Two_models_with_one_id_are_refused_on_add()
+    {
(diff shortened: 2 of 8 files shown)


fsnow:
## Fable review round

Verdict: *"merge-worthy as-is — parts 1 and 2 are correct and part 3 stops the crash"*, with two MED follow-ups that use the PR's own premises as their mechanism. Both are now fixed in `163bdae`, at the seams the review named.

**Confirmed sound** before the findings: `Add`/`Update` are the only routes into `Apply` (both fed by `TypedModels()`, which trims), so the guard covers every draft writer; `AddFromPreset` writes no models; `MarkListing`/`SetModelEnabled` only mutate existing records; settings load constructs nothing new. The **ordinal** predicate is right — every model-id comparison in the AI graph is ordinal (only *connection* ids fold case), so `GPT-4o` and `gpt-4o` are consistently two models and refusing them would have been wrong. `FollowTheChoice` never parks the active pointer on a hidden duplicate.

### 1 (MED) — the padded listing id still misbehaved past the trim

`AiModelCatalog.Parse` took `id` verbatim. Given the padded input the trim fix exists for, the stored `"gpt-4o"` and the fetched `" gpt-4o "` could never join: two rows for one model, `Facts()` never reattaching published facts to the stored row, and — worst — a completed fetch calling `MarkListing` with the padded id, marking the model the reader had just enabled *"no longer listed"*. That's the false alarm #728 exists to avoid. The issue pointed at this seam; the first commit took the narrower one.

Now trimmed in `Parse`, at the one place a listing id enters. A blank id is skipped rather than stored, and the existing listed-vs-parsed count reports it as skipped.

### 2 (MED-LOW) — the tolerance repaired the tab, not the file

`AiModelPickerViewModel.Rebuild` has no dedupe either, and both real routes leave **both** records enabled. Over such a file the picker listed the model twice while `EnableModel`'s `FirstOrDefault` flipped only the first — so switching a model off left its twin on and answering, and the only repair was an edit sheet the new guard now blocks until a row is deleted.

Fixed in `SettingsValidator`, beside the existing no-id sweep — the established repair seam for this class (#319, #787), with its `fixes` log, self-healing on the next save. Trim first, then collapse first-wins; that ordering also retires finding 3 (a hand-edited padded stored id, which the `EnableModel` trim would otherwise have turned into a sibling).

`ApplyFilter` keeps its first-wins tolerance. Nothing should now reach it, but an `ArgumentException` *there* is a Settings-window lockout rather than a bad row.

### Not acted on

**Finding 4 (LOW)** — `SetModelEnabled`'s untrimmed lookup is now inconsistent with `EnableModel`, but it has zero production callers (tests only; the Models tab goes through `EnableModel`). No mechanism reaches it, so no code for it.

### Tests

Two added, both mutation-checked — and the validator test discriminates the trim and the dedupe *independently* (dropping either one fails it):

- `Settings_Sanitize_trims_and_collapses_repeated_model_ids` — the both-enabled shape the review asked for, plus idempotence
- `A_padded_listing_id_is_trimmed`

Full suite: **2628 passed, 0 failed, 17 skipped**.

PR #874, opened 2026-08-28
Cut a passage at an akṣara boundary, not at an arbitrary index (#871)

### LINKED ISSUE #871: Passage truncation is not grapheme-aware: a mid-akṣara cut silently rewrites the boundary word
Found in the 2026-08 Fable code review (`docs/development/CODE_REVIEW_2026-08_FINDINGS.md`, area **R14-2**).

**Severity note:** the reviewer graded this MED on trigger narrowness. I am filing it as high because narrowness is a likelihood argument, not an impact one, and the impact is silently wrong Pāli in the citation path. Downgrade at triage if you disagree — the reasoning is laid out below so the call can be made on the merits.

## What happens

Three character-count cuts in the passage reader stop at an **arbitrary raw Devanagari index**, with no grapheme awareness:

- the hard cap — `WalkForward`, `TeiPassageReader.cs:635` (`rendered >= hardCap → return i`)
- the selection cap — `RawForwardCap`, `:459-482`
- the no-boundary prev-cursor fallback — `WalkBackward`, `:682-685`

Cutting between a consonant and its dependent vowel sign leaves a **bare consonant**. `Deva2Latn` then applies inherent-'a' insertion:

```csharp
// src/CST.Core/Conversion/Deva2Latn.cs:205-209
if (c >= 0x0915 && c <= 0x0939) // consonant: inherent 'a' unless next is a vowel sign or virama
    …
        buf[k++] = 'a';
```

So a window cut inside `…द्ध|ी` ("buddhī") ends **"…buddha"** — a different, entirely plausible Pāli word — and the next page opens on the orphaned mātrā ("ī…"). Stitching consecutive pages does **not** recover it: you get "buddha" + "ī", not "buddhī".

## Why I would call this high

- **It is not visible garbage.** Mojibake announces itself; this produces a well-formed word that is simply the wrong one. Neither a reader nor an agent has any signal that the text was cut.
- **It is in the citation path.** This is the surface an agent quotes the corpus from, and quoted Pāli that is subtly wrong is the worst output this application can produce.
- **Silent wrongness outranks narrow triggers** by the review's own severity bar ("wrong results a reader would act on").

## Triggers

- Any region where the budget + 50% passes with **no danda** — front matter, headings, verse indexes; the same danda-free text the `SentenceScanCap` comment already documents.
- **Any** `SelectionTruncated` cut — `RawForwardCap` counts every rendered char and stops wherever the count lands.

## Fix

After any hard cut, retreat to the previous grapheme-cluster boundary before returning — skip back over combining marks and virama sequences (`StringInfo`, or a manual Mn + virama scan).

## Scope note

The snippet path is **not** affected — its cuts are space/sentence-snapped (verified in review). This is specific to the passage reader's three count-based cuts.

fsnow:
Fixes #871.

Three of the passage reader's cuts are pure character counts that stop wherever the budget runs out. Land one between a consonant and its dependent vowel sign and `Deva2Latn` gives the bare consonant its inherent 'a':

| | rendered |
|---|---|
| source | `ka buddhīkakakaka` |
| cut at the vowel sign | `ka buddha` ← a different, entirely plausible Pāli word |
| next page from that cut | `īkakakaka` ← the orphaned mātrā |

Nothing tells a reader or an agent the text was cut, and stitching the pages back together does **not** recover `buddhī`. This is the surface an agent quotes the corpus from.

### The fix

`ClusterStart` retreats a raw cut to the start of the akṣara it falls inside. Two rules, applied until neither fires: a combining mark **at** the cut belongs to the character before it, and a virama immediately **before** the cut binds to the consonant that follows. Marks are recognised by Unicode category rather than by a Devanagari range, so a conjunct of any depth unwinds without a table to keep current.

**Retreat rather than advance**, because the cut is also the next page's cursor — moving it back puts the whole akṣara at the head of the next window, so paging stitches back to the original text (`"ka bu"` + `"ddhī…"`). Advancing would drop the akṣara from both pages.

One exception: a window narrower than the akṣara it starts on keeps the raw cut. Retreating to nothing would make the window's end its own `NextCursor`, and a client paging forward would never advance — a hang is worse than the boundary letter this fix exists to get right.

### Where it is applied

Two of the three sites are as the issue prescribes: the hard cap in `WalkForward`, and the no-boundary prev-cursor fallback in `WalkBackward`.

**The third is `ExtendToSentenceEnd`'s 4,000-character cap, not `RawForwardCap`.** Every selection end is passed through the sentence extension (`WalkForwardSentences` calls `ExtendToSentenceEnd` unconditionally to finish the selection's own sentence), so the selection clamp's cut never ends a window — where a danda is near, the extension lands on it and there was never a problem. Where none is — front matter, headings, a verse index, the same danda-free text this reader already documents having to cope with — the extension gives up on its 4,000-char cap, and *that* is the raw cut that lands. Aligning `RawForwardCap` would have been unobservable, so it is left alone.

### Tests

Five added, all mutation-checked. Each of the three sites and each of the two rules fails a test independently when reverted:

| mutation | fails |
|---|---|
| hard cap keeps the raw index | `The_hard_cap_…`, `Paging_across_…` |
| prev-cursor fallback keeps the raw index | `The_previous_page_cursor_…` |
| sentence-extension cap keeps the raw index | `The_sentence_extension_cap_…` |
| drop the virama rule only | all four |
| drop the empty-window guard | `A_window_narrower_than_one_aksara_still_advances` |

`Paging_across_a_half_cut_aksara_stitches_back_to_the_word` asserts the whole point: `w1.Text + w2.Text == "ka buddhīkakakaka"`.

Devanagari test data is written as `\uXXXX` escapes per the house rule.

Full suite: **2633 passed, 0 failed, 17 skipped**.

### DIFF
--- a/src/CST.Avalonia.Tests/Search/TeiPassageReaderTests.cs
+++ b/src/CST.Avalonia.Tests/Search/TeiPassageReaderTests.cs
@@ -837,5 +837,137 @@ namespace CST.Avalonia.Tests.Search
             Assert.Equal(5, window.ParagraphNumber);
             Assert.Contains(window.Pages, p => p.Edition == PageEdition.Vri && p.Number == 1);
         }
+    
+        // ---- akṣara-aware cuts (#871) -------------------------------------------------------------------
+
+        /// <summary>
+        /// One Devanagari word in a paragraph with no danda, so the count-based cuts are the only ones that
+        /// can fire. Rendered whole this is "ka buddhīkakakaka"; cut between the conjunct and its vowel sign
+        /// it reads "ka buddha", which is why this bug is worth a test at all.
+        /// </summary>
+        private const string NoDanda =
+            "<body><div id=\"dn1\" type=\"book\">" +
+            "<p rend=\"bodytext\" n=\"5\">" +
+            // ka + space + b-u-d-virama-dh-ii, then four more ka.
+            "\u0915 \u092C\u0941\u0926\u094D\u0927\u0940\u0915\u0915\u0915\u0915</p>" +
+            "</div></body>";
+
+        /// <summary>Where the paragraph's TEXT begins. The paragraph marker points at the opening tag, which
+        /// is zero-width forwards but not a position to do character arithmetic from.</summary>
+        private static int TextStart(string xml) => xml.IndexOf('\u0915');
+
+        /// <summary>
+        /// The hard cap backs off a half-cut akṣara instead of inventing a word.
+        ///
+        /// <para>The budget ran out between the <c>dh</c> (U+0927) and its <c>ī</c> sign (U+0940), leaving a
+        /// bare consonant that <c>Deva2Latn</c> then gave its inherent 'a' — so the window ended "…buddha"
+        /// where the text says "buddhī". Not visible damage: a different, entirely plausible Pāli word, in
+        /// the surface an agent quotes the corpus from.</para>
+        /// </summary>
+        [Fact]
+        public void The_hard_cap_does_not_cut_an_aksara_in_half()
+        {
+            var markers = BookMarkers.Build(NoDanda);
+
+            // maxChars 5 puts the hard cap (maxChars + maxChars/2) exactly on the vowel sign.
+            var w = TeiPassageReader.ReadWindow(NoDanda, TextStart(NoDanda), maxChars: 5,
+                includeVariants: false, outputScript: Script.Latin, markers);
+
+            Assert.Equal("ka bu", w.Text);
+            Assert.DoesNotContain("buddha", w.Text);
+        }
+
+        /// <summary>
+        /// Paging across the cut stitches back to the text, which is the reason the fix retreats rather than
+        /// advances: the window end IS the next window's cursor, so moving it back puts the whole akṣara at
+        /// the head of the next page. Left where the count landed, page two opened on an orphaned vowel sign
+        /// ("īkakakaka") and no amount of stitching recovered "buddhī".
+        /// </summary>
+        [Fact]
+        public void Paging_across_a_half_cut_aksara_stitches_back_to_the_word()
+        {
+            var markers = BookMarkers.Build(NoDanda);
+
+            var w1 = TeiPassageReader.ReadWindow(NoDanda, TextStart(NoDanda), maxChars: 5,
+                includeVariants: false, outputScript: Script.Latin, markers);
+            var w2 = TeiPassageReader.ReadWindow(NoDanda, w1.NextCursor!.Value, maxChars: 50,
+                includeVariants: false, outputScript: Script.Latin, markers);
+
+            Assert.StartsWith("ddhī", w2.Text);
+            Assert.Equal("ka buddhīkakakaka", w1.Text + w2.Text);
+        }
+
+        /// <summary>
+        /// A window narrower than the akṣara it starts on keeps the raw cut rather than coming back empty.
+        ///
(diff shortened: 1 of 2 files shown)


fsnow:
## Fable review round

Three findings, all verified against the actual corpus files rather than argued from shape. All three are fixed in the follow-up commit.

**The judgement call from the first commit was verified, not just accepted.** The reviewer traced `RawForwardCap`'s return to `selectionEnd` (`:168`) and confirmed its only route to a rendered window end is `WalkForwardSentences` → `ExtendToSentenceEnd`, called unconditionally — and that `ExtendToSentenceEnd` can return `from` unchanged only when `from >= limit`, which the budget cannot reach (the div's closing tags cost zero budget and come after the last text char). So aligning `RawForwardCap` would have been unobservable, and `ExtendToSentenceEnd` was the right third site. Worth remembering: if a "no sentence extension" mode is ever added, `RawForwardCap` becomes a live fourth site.

### F1 (MED) — the retreat could step into a tag, leaking markup

Corpus-attested and reproduced empirically. **23 places** carry a niggahita immediately after a close tag (`<hi rend="bold">नेस</hi>` + U+0902), and `e0812n.nrf.xml` opens a paragraph with a vowel sign — the mark's base is on the far side of the markup. Stepping back onto the `>` returned a cursor *inside* the tag, and the previous page rendered `">ṃkaka…"` where pre-PR it rendered `"ṃkaka…"`. A markup leak where there had only been a stranded mark, and a **strict regression** at those points.

Rule 1 now stops at a tag. Not stepping across markup is also what keeps the retreat from moving a cut over a note boundary, so this narrows the rule to what it was already relying on.

### F2 (MED) — a fourth raw cut the issue's list missed

`WalkBackSentences`' scan-floor fallback: where no danda lies within the 2,000-char scan cap behind a selection and no `<p ` opens in the gap, the expansion falls back to the raw floor — opening a window on an orphaned mātrā, the same defect on the selection path. **14 books** carry danda-free runs past that cap; the Paṭṭhāna mūla files are worst (`abh03m5.mul.xml`: a 19,010-char run). Now aligned against the *section* start, so the retreat cannot leave the section.

### F3 (LOW) — rule 2 was blind to virama+ZWJ

**1,083,447 occurrences — 15% of all 7.3M viramas** in the corpus (`पञ्‍ञा`, `सन्‍नि`), concentrated in exactly the danda-free runs where these cuts fire. Without the rule the retreat stopped on the far side of the joiner and a window ended `"pañ"`. Graded LOW because that is a *visible* cut rather than a wrong word, and stitching is lossless — but it made the method's own doc comment false, so it is fixed rather than documented away.

### F4 (LOW) — missing coverage, all four cases added

Including the **anusvara** case, which is the second silent-wrong-word shape: cut before the niggahita, `nesaṃ` reads `nesa` — a real word, no signal. Rule 1 already handled it; nothing pinned it.

### Confirmed correct as written

Termination and bounds on every input; cursor monotonicity and paging termination (the hard-cap guard makes the return strictly `> start`); note-region guarantees at each site, including the ordering of `WalkBackward`'s note-snap *after* the retreat; `SelectionTruncated` accounting; `RenderedLength` vs. the walk's own accounting; the `Materialize` contract. Nukta, U+0951–0954 and ZWNJ are **absent from the corpus** (0 occurrences each, all 217 files scanned) — correctly not handled. No simpler correct version: `StringInfo` grapheme machinery would be heavier and no more correct for this corpus.

### Tests

Four added (nine total for #871), each mutation-checked — reverting any one rule or site fails exactly its own test.

Full suite: **2637 passed, 0 failed, 17 skipped**.

PR #875, opened 2026-08-28
Serve the DPD asset that is actually on disk after an update (#869)

### LINKED ISSUE #869: Pooled SQLite connections serve superseded DPD data after a mid-session asset update (macOS/Linux)
Found in the 2026-08 Fable code review (`docs/development/CODE_REVIEW_2026-08_FINDINGS.md`, areas **R13-1** and **R12-1** — two reviewers reached this independently from the consumer side and the delivery side; it is **one root cause and one fix**).

## What happens

After a mid-session DPD asset **update** on macOS/Linux, the app reports the new asset active while continuing to serve the **superseded** data — in lemma search *and* the dictionary panel — until restart. It can serve a mix of old and new across calls as the pool grows.

1. `DpdUpdateService.InstallFromGzip` (`DpdUpdateService.cs:477`) replaces the live `dpd-cst-subset.db` with `File.Move(overwrite: true)` — a POSIX rename, which silently succeeds over open handles.
2. It raises `AssetInstalled`; `ReopenableLemmaProvider.Reopen` (`ReopenableLemmaProvider.cs:45-53`) builds a fresh `SqliteLemmaProvider`.
3. The new provider uses a **byte-identical connection string**, and Microsoft.Data.Sqlite pools per connection string — so `Open()` hands back pooled handles bound to the replaced file's **old inode**. `Cache=Shared` pins the old pager a second way, independently.
4. The old provider is *retired without* `ClearPool` — it is disposed only at app shutdown.

Log says "Active now" / "Reopened … available=true". Data is stale.

## The rationale in the comment is mistaken

The retire-don't-dispose comment cites an `ObjectDisposedException` race. That does not apply to this provider: its `Dispose` only calls `SqliteConnection.ClearPool`, which by contract leaves in-use connections untouched (they are discarded on close rather than returned to the pool). Disposing the outgoing provider is safe with queries in flight.

## The fix already exists in this repo

`LexiconReader.cs:26-28` documents and fixes this exact footgun with `Pooling=false`, and its comment names the failure — *"pooled handle pointing at the unlinked old inode"*. **DPPN does not have this bug**: `LexiconReader` uses `Pooling=false`, loads fully into memory, and `SqliteDictionarySource` re-stamps and reloads on file change, so mid-session DPPN updates genuinely work.

The fix is to make the DPD path match the sibling already shipping beside it:

- dispose (`ClearPool`) the outgoing provider **before** constructing the replacement in `Reopen`, and
- drop `Cache=Shared` (or use `Pooling=false`, as `LexiconReader` does).

## Scope

- **macOS/Linux only.** Windows is unaffected — the move throws and takes the staged `.pending` path, and `AssetInstalled` is deliberately not fired for a staged install.
- **First install is clean** (#536's case) — there is no pre-existing pool.
- Only an **update over an already-opened** database is affected.
- Ordering is already correct (replace-then-notify, reopen-handler-first), so this is purely connection lifetime — no restructuring needed.

## Why the tests did not catch it

`ReopenableLemmaProviderTests` calls `ClearAllPools()` in its fixture, which masks exactly this.

## Related

`LemmaReportService.InvalidateCache()` has **zero callers** (R13-3) — up to 32 cached dossiers keep serving the old asset's glosses, counts, and `DpdVersion` footer even once this is fixed. Worth doing in the same pass: call it from the `AssetInstalled` binding.

fsnow:
Fixes #869.

On macOS and Linux an in-place asset update is a `File.Move(overwrite: true)` — a POSIX rename, which silently succeeds over open handles. **Two separate things** then went on pointing at the replaced file's old inode, and the app logged the new asset active while answering from the superseded one, in lemma search and the dictionary panel alike, until the next launch.

### 1. The pool — `Reopen` now disposes the outgoing provider *first*

Microsoft.Data.Sqlite pools by connection string, and the replacement provider builds the *same* string, so its very first `Open()` was handed a pooled handle on the old inode. Clearing the pool before constructing the replacement means it opens the file that is actually there.

The retire-don't-dispose this replaces was guarding an `ObjectDisposedException` that cannot happen here: `SqliteLemmaProvider.Dispose` only calls `SqliteConnection.ClearPool`, which by contract leaves in-use connections alone — they are discarded when they close rather than returned — and the instance stays perfectly usable afterwards, merely unpooled. There was nothing to protect, and the protection was what kept the stale handles alive. With nothing retired, the `_retired` list goes with it.

### 2. The shared cache — `Cache=Shared` dropped

A shared cache is keyed by file **path**, so while any connection holds one open, a brand-new connection on that path joins it and reads through the surviving pager. That pinned the old data *independently* of the pool — clearing the pool does not touch it — and the two tests below prove them apart. It bought nothing here either: connections are short-lived and read-only, and shared cache adds cross-connection table locking between the GUI and the API for the privilege.

This is what `LexiconReader` already documents and avoids (`Pooling=false`, *"pooled handle pointing at the unlinked old inode"*), which is why DPPN never had this bug.

### 3. `InvalidateCache`, folded in per the issue (R13-3)

It had **no callers anywhere in the tree**, so up to 32 assembled dossiers kept serving the old asset's glosses, corpus counts and `DpdVersion` footer for every lemma the reader had already visited — the same staleness one layer up, and it would have survived the fix below it untouched. Now called from the same `AssetInstalled` handler, for the dpd asset only.

### Why the tests did not catch it

`ReopenableLemmaProviderTests`' fixture called `SqliteConnection.ClearAllPools()` unconditionally — precisely the state the app is never in. Four tests passed and the bug shipped. The clear is now opt-out and the new tests pass `false`.

### Tests

Three added, each mutation-checked — and the two mechanisms discriminate **independently**:

| mutation | fails |
|---|---|
| `Reopen` doesn't dispose the outgoing provider | both replacement tests |
| put `Cache=Shared` back | only `A_shared_cache_connection_held_open_…` |
| drop the `InvalidateCache` call | only the report-cache test |

`An_asset_replaced_mid_session_is_served_from_the_new_file` does the real sequence: resolve a form (pooling a handle), build the replacement beside it, `File.Move` over the live file, `Reopen`, then assert the new form resolves **and the old one no longer does**.

Full suite: **2640 passed, 0 failed, 17 skipped**.

Windows is unaffected either way — the move throws there and takes the staged `.pending` path, which deliberately does not fire `AssetInstalled`.

### DIFF
--- a/src/CST.Avalonia.Tests/Services/FirstRunDictionaryVisibilityTests.cs
+++ b/src/CST.Avalonia.Tests/Services/FirstRunDictionaryVisibilityTests.cs
@@ -175,6 +175,42 @@ public sealed class FirstRunDictionaryVisibilityTests : IDisposable
         Assert.Contains("dpd", PickerIds(prefs));
     }
 
+    /// <summary>
+    /// The dossier cache is cleared alongside the reopen.
+    ///
+    /// <para>A lemma report is assembled from the asset's glosses and corpus counts and stamped with its
+    /// version, and up to 32 are held. Reopening the provider without clearing them left the reader looking
+    /// at the superseded asset's report — footer version and all — for every lemma they had already visited:
+    /// the same staleness as #869 one layer up, and it would have survived that fix untouched.
+    /// <c>InvalidateCache</c> was written for this and had no callers anywhere in the tree. (#869)</para>
+    /// </summary>
+    [Fact]
+    public void The_bound_handler_clears_the_report_cache_when_the_dpd_asset_lands()
+    {
+        var (lemma, _) = BuildApp();
+        var updates = new FakeUpdates();
+        var reports = new CountingReports();
+        Assert.True(App.BindLemmaReopen(updates, lemma, reports));
+
+        BuildDpdAsset(_dpdPath);
+        updates.RaiseAssetInstalled(App.DpdAssetId);
+
+        Assert.Equal(1, reports.Invalidations);
+
+        // And not for somebody else's asset, which does not reopen the provider either.
+        updates.RaiseAssetInstalled("dppn");
+        Assert.Equal(1, reports.Invalidations);
+    }
+
+    private sealed class CountingReports : ILemmaReportService
+    {
+        public int Invalidations { get; private set; }
+        public bool IsAvailable => true;
+        public void InvalidateCache() => Invalidations++;
+        public Task<LemmaReport?> BuildAsync(long lemmaId, CancellationToken ct = default) =>
+            Task.FromResult<LemmaReport?>(null);
+    }
+
     // A lexicon install must not reopen the lemma provider — it is a different asset, and reopening on every
     // event would retire a live inner provider for nothing.
     [Fact]
--- a/src/CST.Avalonia.Tests/Services/LemmaReportCacheTests.cs
+++ b/src/CST.Avalonia.Tests/Services/LemmaReportCacheTests.cs
@@ -1,0 +1,77 @@
+using System.Threading;
+using System.Threading.Tasks;
+using CST.Avalonia.Models;
+using CST.Avalonia.Services;
+using CST.Conversion;
+using CST.Lemma;
+using Moq;
+using Xunit;
+
+namespace CST.Avalonia.Tests.Services;
+
+/// <summary>
+/// What the dossier cache is allowed to keep across an asset swap. (#869)
+/// </summary>
+public class LemmaReportCacheTests
+{
+    private static DpdLemmaMeta MetaFor(string version) =>
+        new("full", version, "3", "1", "CC BY-NC-SA", "DPD", "dpd", "Bodhirāsa", "https://example.invalid");
+
+    private static LemmaDetail Detail() =>
+        new(1, "dhamma", "nt", "a test gloss", null, null, null, null, null, null,
+            null, null, null, null, null, null);
+
+    /// <summary>
+    /// A report assembled from the asset that was replaced while it was being assembled is handed to the
+    /// caller that asked for it, and NOT put in the cache.
+    ///
+    /// <para>Assembly awaits several corpus searches — seconds, which is the whole reason this cache exists —
+    /// so an install landing inside that window clears the cache before the report arrives. Storing it
+    /// anyway put a dossier built from the superseded asset back into an otherwise clean cache, version
+    /// footer and all, where it sat until 32 other lemmas evicted it. The trigger is ordinary: the reader
+    /// opens a lemma report while the background update installs.</para>
+    /// </summary>
+    [Fact]
+    public async Task A_report_assembled_across_an_asset_swap_is_not_cached()
+    {
+        var meta = MetaFor("v0.4.20260531");
+        var lemma = new Mock<ILemmaProvider>();
+        lemma.SetupGet(l => l.IsAvailable).Returns(true);
+        lemma.SetupGet(l => l.Meta).Returns(() => meta);
+        lemma.Setup(l => l.GetDetail(1)).Returns(Detail());
+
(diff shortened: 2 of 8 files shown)


fsnow:
## Fable review round

**The fix is confirmed correct on the platforms where the bug exists.** The reviewer traced the install path end to end and could not construct a surviving route to stale data, and — the part I most wanted checked — verified the concurrency claim against Microsoft.Data.Sqlite 10.0.9's actual pool semantics rather than my reading of the contract:

> `Clear()` disposes idle pooled connections and marks checked-out ones do-not-pool, so a query in flight **cannot repopulate the pool with a stale handle** — its pre-rename connection is discarded on return, not repooled.

Enumerated at the swap: idle-pooled → closed by Clear; busy at Clear → discarded on return; opened on the old inner *after* Clear → opens the post-rename file, harmless. Dropping `_retired` is not a leak (the old inner's entire teardown *is* `ClearPool`, and in-flight callers keep it alive on their stacks). Concurrent `Reopen`s are benign. `ClearPool` covers every connection-string variant in play — there is exactly one per path, and `ReadDpdVersion` builds the identical string and clears its own.

Also confirmed: `Cache=Shared` was the **only** shared-cache user in the tree, so removing it regresses nothing; the shared-cache test is legitimate rather than accidentally passing (SQLite matches the shared-cache lookup by canonical path); and the `InvalidateCache` wiring reaches both GUI and API, since `LocalApiServer` resolves the same singleton from the same container.

### F1 (HIGH) — the two new tests would have failed on Windows

Both rename over a file with open SQLite handles, and SQLite's win32 VFS opens without `FILE_SHARE_DELETE` — so the move throws. That is *exactly* why `InstallFromGzip` stages a `.pending` file there and does not raise `AssetInstalled` at all. Windows is a supported test platform in this repo.

Gated with `UnixFact`, whose skip reason is now a constructor parameter: the shell probe and a POSIX rename are not the same reason to skip, and a skip that explains the wrong one costs the reader the trip to find out which. The gating is semantically right too — on Windows the mechanism under test does not exist. Verified by inverting the platform check that they **skip** rather than silently pass, and that the existing `UnixFact` shell-probe tests still run.

### F2 (LOW) — a report assembled *across* the swap could be re-cached

Assembly awaits several corpus searches (seconds — the reason the cache exists), so an install landing inside that window clears the cache *before* the report arrives, and storing it anyway put a dossier built from the superseded asset back into an otherwise clean cache — version footer and all — until 32 other lemmas evicted it. Concrete trigger: the reader opens an uncached lemma report while the background update installs.

Now compares the asset version across assembly. The caller still gets what it asked for; it is simply not kept. Two tests added (`LemmaReportCacheTests`), including one that the ordinary case *is* still cached — otherwise the fix would be a cache that never holds anything.

### F3 (LOW) — nothing pinned the production `reports` resolution

A DI registration renamed away would have dropped the invalidation silently. `SubscribeLemmaReopen` now logs an error when `ILemmaReportService` is not registered — the same loudness the provider half already has (`#563`), for the same reason.

### F4 — not acted on

`Pooling=false` throughout (the `LexiconReader` pattern) is a viable simpler variant with an extra payoff: our own constructor-created pooled handle is what forces Windows into the staged path, so `Pooling=false` would likely make Windows mid-session updates go live in place. Cost is a connection open + schema parse per call on the N+1 `GetDetail` dictionary-lookup path. **Design context, not a defect** — the fix as written is correct and keeps the hot path warm. Worth its own issue if Windows mid-session liveness becomes a goal.

The reviewer also declined to file one window under scope discipline, which I'd rather record than lose: a caller holding the old inner across the swap runs one query against the new file with the old instance's schema flags. That only misbehaves if an update *removes* columns, and no producer of a schema-shrinking asset exists.

Full suite: **2642 passed, 0 failed, 17 skipped** (macOS; two more skip on Windows by design).

Issue #803, opened 2026-08-21
Keep what parses: one bad property in settings.json should not cost the whole file


fsnow:
Split out of #785, whose headline — settings had no backup recovery — is fixed by #799. This is the other half named there, and it is a different change with a different argument, so it gets its own issue rather than holding that one open.

## What still happens

`SettingsService.LoadSettingsAsync` deserializes the whole file in one call. Any single property `System.Text.Json` cannot convert throws, and the catch treats the entire file as unusable:

```
JsonException: The JSON value could not be converted to List<AiHeaderRecord>.
Path: $.Ai.Chat.Connections[0].Headers
```

That one property took the XML books directory, fonts, layout, every AI connection and every model list with it.

**#799 makes that survivable, not correct.** Recovery restores the previous save — so the reader keeps their configuration minus whatever changed since. Per-section tolerance would have made this incident cost a *header*.

## Shape

Undecided, and worth arguing rather than reaching for the obvious.

The blunt version — deserialize each top-level section independently and default only the ones that throw — is more invasive than it sounds: `Settings` is one record, and splitting its load means either a hand-written reader or a converter per section, both of which add a place for the shapes to drift apart. It also raises a question nobody has answered: **when a section is dropped, is the reader told?** A silently defaulted `Ai` section is the same defect this issue is about, one level down — the file loaded, nothing said, and the connections are gone.

There is a cheaper variant worth weighing first: tolerate at the level where the damage actually concentrates. Every real instance so far has been *inside* `Ai.Chat.Connections` — a header shape, a null collection, a null entry. A converter that skips an unreadable **connection** rather than failing the file would have covered the observed incident and both near-misses, without restructuring the load.

## Why it matters beyond settings

`ApplicationState` has the identical structure and the identical catch. It is better protected only because it has had backups for longer, and backups share the primary file's shape — so a persisted-type change breaks them all at once (#787).

## Related

- #785 — backup recovery, fixed in #799
- #787 — does this build load what the previous build wrote; fixed the null-collection family that would otherwise arrive here
- #319 — a null section bricking the Settings window: the same shape, from the other direction

Issue #785, opened 2026-08-21
Settings has no backup recovery, though ApplicationState already does — a bad property destroys the file


fsnow:
Found by the incident #784 fixed. That PR fixes the *cause* — a header shape the previous version could not read. This is about why one unreadable property destroyed a whole configuration, and it turns out the app already knows how to prevent that.

## Two persisted stores, opposite treatment

**`ApplicationStateService` survives a corrupt file automatically.** It writes a timestamped backup on every save, and on a parse failure walks its backups newest-first until one deserializes, then migrates and sanitizes it exactly like the primary file (`ApplicationStateService.cs:178`, `:505`). Retention is 8 recent plus 14 daily (`:427-434`).

**`SettingsService` has none of it.** It catches, replaces the settings object with defaults, and calls `RequestSave()`:

```csharp
catch (Exception ex)
{
    _logger.Error(ex, "Failed to load settings from {Path} - reverting to defaults", _settingsFilePath);
    _settings = new Settings();
    ApplyFirstRunDefaults();
    RequestSave();
}
```

That `RequestSave()` is what makes it unrecoverable. The unreadable file is **overwritten with the defaults** — so the configuration is not unavailable for the session, it is gone. In the incident that prompted this, a hand-built model list was rebuilt by hand because there was nothing left on disk.

One property `System.Text.Json` could not convert cost the XML books directory, fonts, layout, every AI connection and every model list.

## The fix is the pattern that already exists

**[fsnow]** ruled out the obvious alternative, and correctly: moving the bad file aside as `settings.json.unreadable` is a fix for whoever wrote the code, not for a reader — nobody would ever find it, and by the time they thought to look they would have rebuilt their configuration already.

Automatic recovery from a backup has none of that problem, because the reader never learns anything went wrong. The previous save simply loads. And it needs no new design: `ApplicationStateService` has the write path, the read-newest-first path, the retention policy and the tests (`BackupRetentionTests`).

So: **give `SettingsService` the same backup-and-recover treatment.** Whether the machinery is shared or mirrored is an implementation question; the behaviour should be identical, because the reason for it is identical.

Until a backup exists to fall back to, the catch should also **stop overwriting** the file it could not read — defaulting in memory is what STATE-3's comment above that catch actually argues for, and nothing in that reasoning requires destroying the previous file.

## Per-section tolerance, separately

[...]

PR #884, opened 2026-08-29
Give application-state.json the recovery its settings sibling has (#877, #879)

### LINKED ISSUE #877: application-state.json has none of settings.json's recovery protections, and its own backups can lock in defaults
Found in the 2026-08 Fable code review (`docs/development/CODE_REVIEW_2026-08_FINDINGS.md`, **R11-2**), verified against current source.

`settings.json` was hardened three times (#785, #787, #803). Its sibling `application-state.json` — open books, reading positions, tree expansion, window geometry — received none of it. Three distinct gaps, one file:

## (a) An unreadable state file is overwritten, not preserved

`LoadStateAsync`'s catch installs defaults (`ApplicationStateService.cs:187-196`) and nothing copies the unreadable file aside. The next save — the 60s timer, or shutdown `ForceSaveAsync`, which saves **unconditionally** — writes over the only copy. `SettingsService.PreserveUnreadable()` exists for exactly this.

## (b) The defaults save becomes the newest backup, permanently

`SaveStateAsync` (`:245`) calls `WriteBackupAsync` before **every** save, with no equivalent of `SettingsService._backupNextSave`. So:

1. state file becomes unreadable (a persisted-type change breaks it — the expected trigger)
2. defaults are installed and saved → **defaults are now the newest backup**
3. `TryLoadFromBackupAsync` returns on the **first** backup that deserializes

The reader's real session sits one file below and is never reached, on this launch or any future one. `SettingsService` guards this and names the trigger in a comment; ApplicationState reproduces it.

## (c) An empty state file skips the backups entirely

`state == null` at `:160` returns `false` from inside the `try`, so the catch never runs and `TryLoadFromBackupAsync` is never called. An empty/whitespace file gets defaults with backups sitting right there.

## Fix

Mirror `SettingsService`: preserve-aside on an unreadable file, a `_backupNextSave = false` equivalent for the defaults save, and route the null case through the backup walk.

## Scope

Loss is the session, not hand-built config — which is why the reviewer graded it MED. Filing as high because (b) is **not** a one-time loss: once a defaults backup is newest it wins on every subsequent launch, and nothing in the UI says so.

### LINKED ISSUE #879: A state change landing mid-save is silently dropped until the next one
Found in the 2026-08 Fable code review (**R11-3**), verified against current source.

Lost-update race on the dirty flag. `SaveStateAsync` serializes its snapshot early (`ApplicationStateService.cs:225`), and the callers clear `_isDirty` only **after** the save succeeds (`:136`, `:570`). A `MarkDirty()` landing between the serialize and the clear is wiped without its change being in the file — and the change then stays unsaved until the *next* `MarkDirty()`.

Graceful shutdown is covered: `ForceSaveAsync` saves unconditionally and the shutdown path re-captures VM state first. So the loss needs a crash or force-quit inside that window — bounded and rare.

## Fix

Clear the flag when the snapshot is taken, and re-set it on failure. One line each side.

fsnow:
Fixes #877. Fixes #879.

First of two PRs for the Settings/application-state hardening pass. `settings.json` was hardened three times (#785, #787, #803); `application-state.json` — open books, reading positions, tree expansion — received none of it. It also had **no test seam**, so nothing could reach its load path to notice. The asymmetry in the tests mirrored the asymmetry in the code, and hid it.

### #877(a) — an unreadable state file is now kept

Nothing kept it before. The next save — the 60s timer, or shutdown `ForceSaveAsync`, which saves *unconditionally* — wrote over the only copy, and the session was gone with no way back even by hand. Now moved aside as `application-state.unreadable-<timestamp>.json`, the way `SettingsService.PreserveUnreadable` has done since #785.

### #877(b) — the defaults save no longer becomes the newest backup

This is the one with teeth. A backup was written before **every** save, and `TryLoadFromBackupAsync` stops at the first file that deserializes — so defaults became the *newest* backup and won every subsequent recovery, with the reader's real session one file below, perfectly readable, never reached. The expected trigger is a persisted-type change, which breaks the primary file and the recent backups together, so this is the ordinary case rather than a corner of it.

`_backupNextSave` mirrors `SettingsService`, including that it suppresses exactly **one** save — a permanent stop would silently end the reader's backup history.

### #879 — a change made during a save is no longer discarded

`SaveStateAsync` serializes its snapshot before the first await, and the callers cleared the dirty flag only after the write returned. A `MarkDirty` landing in between was wiped: not in the snapshot, no longer marked, and unsaved until something else happened to mark the state dirty again. A crash or force-quit in that window lost it.

The flag is now cleared **when the snapshot is taken** and re-set if the write fails — and the two callers no longer clear it, which would have undone exactly this fix.

### A correction to the finding

The review's third sub-point — *"`state == null` (empty file) returns without even trying the backups"* — **is wrong**, and I'd rather say so than quietly fix past it. `Deserialize("")` and `("   ")` both throw (`The input does not contain any JSON tokens`), so an empty file has always been caught and recovered. Only a file holding the literal token `null` reaches that branch, and nothing in the app writes one.

I kept the change — that branch gave up without trying the backups, which no other unreadable file gets, so routing it through the same recovery leaves one behaviour instead of two — but as a **simplification, not a fix**, and the test says so. Both the empty-file case and the literal-`null` case are now pinned, so the next reader of that document finds the claim already settled. Recorded on #877 as well.

### Tests

Nine, in a new `ApplicationStateRecoveryTests` — the first tests to exercise this service's load path at all. Each of the four changes mutation-checked independently:

| mutation | fails |
|---|---|
| drop the preserve-aside | `An_unreadable_state_file_is_kept_beside_the_defaults` |
| back up unconditionally again | `The_defaults_save_after_an_unreadable_file_leaves_no_backup` |
| null branch returns without recovering | `A_state_file_holding_literal_null_is_recovered_from_a_backup` |
| clear the dirty flag after the write, as before | `A_change_made_during_a_save_leaves_the_state_dirty` |

The #879 test reads like a race but is deterministic: the snapshot and the clear both happen synchronously, before `ForceSaveAsync` can return its task, so the `MarkDirty` is always the mid-save one.

Also pinned: the save *after* the defaults backs up again, and an ordinary save still clears the flag — otherwise each fix would be a cache that never holds anything and a state that saves forever.

Full suite: **2651 passed, 0 failed, 17 skipped**.

### DIFF
--- a/src/CST.Avalonia.Tests/Services/ApplicationStateRecoveryTests.cs
+++ b/src/CST.Avalonia.Tests/Services/ApplicationStateRecoveryTests.cs
@@ -1,0 +1,223 @@
+using System;
+using System.IO;
+using System.Linq;
+using System.Threading.Tasks;
+using CST.Avalonia.Models;
+using CST.Avalonia.Services;
+using Microsoft.Extensions.Logging.Abstractions;
+using Xunit;
+
+namespace CST.Avalonia.Tests.Services;
+
+/// <summary>
+/// What happens to <c>application-state.json</c> when it cannot be read. (#877, #879)
+///
+/// <para>The settings side has had this since #785 — preserve the unreadable file, walk the backups, and
+/// make the defaults save leave no backup — and its sibling never received any of it. The loss is a session
+/// rather than hand-built configuration, which is why it took longer to matter, not why it does not.</para>
+///
+/// <para>These are the first tests to exercise this service's load path at all: it had no directory seam, so
+/// the existing state tests could only reach the validators. The asymmetry in the tests mirrored the
+/// asymmetry in the code, and hid it.</para>
+/// </summary>
+public sealed class ApplicationStateRecoveryTests : IDisposable
+{
+    private readonly string _dir;
+    private readonly string _statePath;
+    private readonly string _backupDir;
+
+    public ApplicationStateRecoveryTests()
+    {
+        _dir = Path.Combine(Path.GetTempPath(), $"app-state-recovery-{Guid.NewGuid():N}");
+        Directory.CreateDirectory(_dir);
+        _statePath = Path.Combine(_dir, "application-state.json");
+        _backupDir = Path.Combine(_dir, "app-state-backups");
+    }
+
+    private ApplicationStateService Service() =>
+        new(NullLogger<ApplicationStateService>.Instance, _dir);
+
+    private string[] Backups() =>
+        Directory.Exists(_backupDir) ? Directory.GetFiles(_backupDir, "application-state-*.json") : [];
+
+    /// <summary>A state file naming one open book, so a restore can be told from defaults.</summary>
+    private static string StateNaming(string book) =>
+        $$"""
+        { "version": "1.0", "bookWindows": [ { "bookFileName": "{{book}}", "bookIndex": 1 } ] }
+        """;
+
+    private static string? OpenBook(ApplicationState state) =>
+        state.BookWindows.FirstOrDefault()?.BookFileName;
+
+    // ---- the unreadable file itself (#877a) ------------------------------------------------------------
+
+    /// <summary>
+    /// A state file that cannot be read is kept, not overwritten.
+    ///
+    /// <para>Nothing else keeps it: the very next save — the 60s timer, or the shutdown
+    /// <c>ForceSaveAsync</c>, which saves unconditionally — writes over the only copy, and the reader's
+    /// session is gone with no way back even by hand.</para>
+    /// </summary>
+    [Fact]
+    public async Task An_unreadable_state_file_is_kept_beside_the_defaults()
+    {
+        File.WriteAllText(_statePath, "{ this is not json");
+
+        using var service = Service();
+        await service.LoadStateAsync();
+
+        var kept = Directory.GetFiles(_dir, "application-state.unreadable-*.json");
+        Assert.Single(kept);
+        Assert.Contains("this is not json", File.ReadAllText(kept[0]));
+    }
+
+    // ---- the defaults save must not become the newest backup (#877b) -----------------------------------
+
+    /// <summary>
+    /// The defaults installed after an unreadable file leave no backup behind.
+    ///
+    /// <para>A backup is written before <b>every</b> save, and <c>TryLoadFromBackupAsync</c> stops at the
+    /// first file that deserializes — so without this the defaults become the newest backup and win every
+    /// subsequent recovery, with the reader's real session one file below, perfectly readable, never
+    /// reached. The expected trigger is a persisted-type change, which breaks the primary file and the
+    /// recent backups together, so this is the ordinary case rather than a corner of it.</para>
+    /// </summary>
+    [Fact]
+    public async Task The_defaults_save_after_an_unreadable_file_leaves_no_backup()
+    {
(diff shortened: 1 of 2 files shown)


fsnow:
## Fable review round

The reviewer confirmed the flag lifecycle, the dirty-flag repositioning, the seam split, and the determinism of the mid-save test — and independently verified my correction to the finding (empty/whitespace **throws**; only literal `null` reaches that branch; nothing writes one). But it found that **the port was incomplete in a way that made one path worse than `main`**, which is the right call and the reason to run these.

### MED-1 — preserve-aside opens a window I didn't close

Moving the primary file away is what *creates* the exposure. On `main` the unreadable file stayed put, so a crash after a recovery simply meant the next launch ran recovery again and found the same backup. With preserve-aside and nothing else:

1. unreadable primary → moved aside → backup restored **in memory only**; the restore does not `MarkDirty`, so *nothing was scheduled to write it at all*
2. crash inside that window (in practice ~the first minute, until something else marks the state dirty)
3. next launch: no primary file → **"first run"**, backups never consulted → session lost a second time
4. and that process's first save writes **defaults as the newest backup** — reintroducing exactly the poison #877(b) exists to prevent

`SettingsService` states the fix order in its own comments (*"a restore that has to survive a race to count is not a restore"*). Both pieces now ported:

- **the write-back**, awaited, inside the restore path — `_backupNextSave` is correctly still `true` there, since a restored session *should* become the newest backup; only defaults must not
- **the no-file backup check**, as the belt to that suspender

### LOW-1 — the comment shipped the claim my own test refutes

The branch comment and log said "empty or whitespace". Corrected to say what the branch actually catches and why the change is a simplification rather than a fix.

### LOW-3 — `Dispose` no longer drained an in-flight save

A regression this PR introduced. `Dispose` used to drain by *accident*: the flag stayed set during a write, so it saw dirty, started a redundant second save, and that save's `Wait` blocked on the semaphore. With the flag now cleared at the snapshot, an in-flight write no longer looks dirty and `Dispose` returned while the process tore down under it. Now drains the semaphore explicitly. Bounded either way (temp + `File.Replace`: an interrupted write loses that snapshot, never the file), but the behaviour was real and shouldn't vanish as a side effect of the #879 fix.

### LOW-2 — not acted on

Clearing the dirty flag after `Dispatcher.UIThread.Invoke` returns leaves an instruction-scale residue of #879 (old window: an entire disk write). Closing it fully means clearing *inside* the marshalled serialize lambda — giving `SerializeCurrent` a side effect its name disclaims — and no test can discriminate it. The reviewer said it would not block on this; I agree, and would rather leave the window documented than trade it for a method that lies about what it does.

### Tests

Four added (13 total), each mutation-checked. One of them covers a mutation the reviewer found slipping through all nine originals:

| mutation | fails |
|---|---|
| no write-back after a restore | `A_restore_is_written_back_to_the_primary_file_at_once`, `The_write_back_after_a_restore_does_leave_a_backup` |
| no-file branch skips the backups | `A_missing_state_file_with_backups_present_is_recovered_not_treated_as_a_first_run` |
| hoist `_backupNextSave = false` above the backup walk | `The_write_back_after_a_restore_does_leave_a_backup` |

Plus `A_genuine_first_run_is_still_a_first_run`, so the new guard can't turn an empty backup directory into a recovery that reports something it didn't find.

Full suite: **2655 passed, 0 failed, 17 skipped**.

PR #885, opened 2026-08-29
Serialize settings writes, and close four ways a setting could vanish (#878, #880–#883)

### LINKED ISSUE #878: Concurrent settings saves are unserialized and share one temp path
Found in the 2026-08 Fable code review (**R11-1**), verified against current source.

`SettingsService.SaveSettingsAsync` (`:402-441`) has no write serialization — the `_saveLock` in that file guards only the debounce flag, not the write. `ApplicationStateService` fixed exactly this defect with a `SemaphoreSlim(1,1)` (STATE-2); settings never got it.

**Concurrent callers exist today:** the 750ms debounce timer flushes on a pool thread while `XmlUpdateService.cs:226` and `IndexingService.cs:62` call `SaveSettingsAsync()` directly from background flows.

## The interleaving

Both writers use the same `settings.json.tmp`:

1. A finishes `WriteAllTextAsync(tmp)`
2. B truncates and starts rewriting the same tmp
3. A's `File.Replace` promotes B's **half-written** tmp over `settings.json`

On Windows one writer instead takes an `IOException` and that save is silently lost.

Relatedly, shutdown `FlushPendingSaveAsync` reads `_savePending == false` while a timer-triggered save is still mid-write, and lets the process exit under it.

## Fix

A `SemaphoreSlim(1,1)` around the write, as in `ApplicationStateService.SaveStateAsync`. Note the name collision: the existing `_saveLock` is a plain `object` for the debounce flag, so the write semaphore needs its own name.

The tolerant reader and backup walk make permanent loss unlikely — but this is the save path writing a corrupt file, and the recovery machinery should not be load-bearing for an ordinary save.

### LINKED ISSUE #880: The tolerant settings reader ignores property-level [JsonConverter], so salvage is weaker than a strict read
Found in the 2026-08 Fable code review (**R11-4**), verified against current source.

`SalvageObject` (`TolerantSettingsReader.cs:167`) recurses with the property's **type** only:

```csharp
var value = Node(json, property.PropertyType, $"{path}.{property.Name}", options, dropped);
```

That loses any `[JsonConverter]` declared on the property. Concretely: `AiConnectionRecord.Headers` carries `[JsonConverter(typeof(AiHeaderRecordListConverter))]` (`Settings.cs:223`), which reads the legacy dict-shaped form fine on the strict path. Taken apart property-by-property, `Node` sees `List<AiHeaderRecord>`, the JSON is an object, `SalvageList` does not apply and `DictionaryValueType` is null — so it falls to the drop-the-collection branch (`:88-92`) and the headers are discarded.

Nothing is silently lost (the drop is named in the log and the original is copy-preserved), but **the salvage path is strictly weaker than the strict path for any property carrying its own converter** — which is a trap for the next such converter, in the one mechanism whose whole job is to lose as little as possible.

## Fix

In `SalvageObject`, when a property declares a `[JsonConverter]`, attempt the whole-node strict read with that converter before falling through to `Node`.

### LINKED ISSUE #881: Sanitize never re-seeds a missing script-font key, so a script can vanish from the Appearance panel for good
Found in the 2026-08 Fable code review (**R11-5**), verified against current source.

`SettingsValidator.Sanitize` removes unknown `ScriptFonts` keys and repairs null/bad values, but never adds back a **missing** canonical key (`SettingsValidator.cs:184-205`).

The Appearance panel builds its rows by enumerating that dictionary (`SettingsViewModel.cs:273`), so a script entry dropped by a tolerant salvage or a hand-edit leaves that script **permanently absent from the fonts UI**, with no way back short of editing `settings.json` by hand. Rendering itself falls back safely (`BookZoomService.cs:126-130` seeds on write; `BookFontResolver`/`FontService` default on a miss) — so nothing looks broken, the control is simply not there.

Compounding it: the salvage's Error message (`SettingsService.cs:206-208`) says dropped nodes "were reset to their defaults", which for a dictionary entry is untrue — it is removed, and stays removed.

## Fix

Have `Sanitize` add any of the canonical script keys that are absent, with the same defaults `FontSettings`' constructor uses.

### LINKED ISSUE #882: Changing the log level silently kills file logging for the session (Windows)
Found in the 2026-08 Fable code review (**R11-6**). Graded ❓ by the reviewer — the code shape is confirmed, the runtime effect was not reproducible on macOS.

`SettingsViewModel.ReconfigureLogger` (`:1745-1785`) assigns a brand-new `Log.Logger` without disposing or `CloseAndFlush`-ing the old one.

The old pipeline's rolling file sink keeps today's log file open. Serilog's file sink defaults to exclusive access (`shared: false`), so on Windows the new logger's file sink cannot open the same file: **file logging silently dies for the rest of the session the moment the reader changes the log level** — and the Logging panel appears to work, because the console sink still logs.

On macOS/Linux there is no exclusive lock, which is why development never sees it.

## Fix

Capture the old logger and dispose it after the swap.

## Verification

Needs a Windows run (Placid or Kingfisher): change the log level in Settings, then check whether anything further is written to `logs/cst-avalonia-*.log`.

### LINKED ISSUE #883: An older build strips a newer build's settings on launch, before the reader touches anything
Found in the 2026-08 Fable code review (**R11-7**), verified against current source.

Neither `Settings` nor `ApplicationState` carries a `[JsonExtensionData]` overflow member, so any property a newer build added is silently stripped on the first round-trip through an older build.

For settings it is worse than a round-trip: the *"version X is newer than supported; reading as-is"* note (`SettingsValidator.cs:44-47`) counts toward `notes.Count > 0`, which triggers an immediate `RequestSave()` (`SettingsService.cs:93-95`) — so merely **launching** the older build rewrites the file in the older shape, before the reader changes anything. The comment says "reading as-is"; the code then writes it back reduced.

Backups written by the newer build mitigate, but recovery is manual and nothing says it happened.

## Fix options

1. Do not save on the newer-version note alone (the cheap half — it makes "reading as-is" true).
2. Add a `[JsonExtensionData]` member so unknown properties round-trip (the real fix).

Worth doing both: (1) stops the launch-time rewrite, (2) stops the loss whenever a save does happen.

fsnow:
Fixes #878. Fixes #880. Fixes #881. Fixes #882. Fixes #883.

Second of two PRs for the Settings/application-state hardening pass (#884 was the state half). Five findings, one subsystem.

### #878 — concurrent saves were unserialized and shared one temp path

`SaveSettingsAsync` had no write lock; the `_saveLock` beside it guards the debounce **flag** and nothing else. Concurrent callers are ordinary, not hypothetical — the 750ms debounce flushes on a pool thread while `XmlUpdateService` and `IndexingService` save directly from background flows. Both writers used one `settings.json.tmp`, so A could finish writing the temp file, B truncate and start rewriting it, and A's `File.Replace` promote B's half-written temp over the real file. `ApplicationStateService` has serialized its writes since STATE-2.

`FlushPendingSaveAsync` now drains too: shutdown awaits it, and it returned while a save started elsewhere was still mid-write.

### #880 — the salvage path ignored a property's own `[JsonConverter]`

`Node()` takes a `Type`, so an attribute living on the **property** was lost the moment we recursed. `AiConnectionRecord.Headers` is the live case: `AiHeaderRecordListConverter` reads the legacy dict-shaped form fine on the strict path, but taken apart property-by-property the node is an object where a `List` belongs, which lands in the drop-the-collection branch. **The salvage path was strictly weaker than the strict path** — in the one mechanism whose whole job is to lose as little as possible.

Worth noting what the test had to do to reach it: the reader tries a whole-node read at every level first (deliberately, so ordinary converters apply wherever they can), so the gap only shows once something *else on the same object* has failed. My first fixture broke the file further up the tree, the `ai` node deserialized whole, and the test passed with the fix reverted.

### #881 — a script could vanish from the Appearance panel for good

The panel builds its rows by enumerating `ScriptFonts`, so a key dropped by a salvage or a hand-edit left that script with **no font control at all** — permanently, since nothing re-seeded it, and invisibly, since rendering falls back safely.

Seeded from a new `FontSettings.DefaultScriptFonts()` that the constructor now uses too, so the two lists cannot drift. A second hand-written list would have been the same defect one level up.

### #882 — changing the log level killed file logging (Windows)

The outgoing pipeline's rolling file sink keeps today's log file open, and Serilog's file sink defaults to exclusive access — so on Windows the incoming logger's file sink cannot open the same file and file logging dies silently for the rest of the session, with the Logging panel still looking like it worked because the console sink is unaffected.

Disposing the replaced logger releases the handle. **After the swap, never before** — Serilog answers a missing logger with a silent no-op rather than an error, so disposing first would leave a window where anything logged simply does not exist. Split into `SwapGlobalLogger` so a test can pin that ordering; both halves are mutation-checked.

Still wants a Windows run to confirm the symptom is gone (Placid or Kingfisher): change the log level, then check whether anything further reaches `logs/cst-avalonia-*.log`.

### #883 — an older build stripped a newer build's file on launch

Both stores gained a `[JsonExtensionData]` member so unknown **top-level** properties round-trip. Nested sections still shed; that is the cheap half taken deliberately, and the doc comment says so rather than implying more.

And neither store writes a newer-than-supported file back on load. *"Reading as-is"* was true of the read and false of what followed: the note counted toward "something changed", so merely **launching** an older build rewrote the file in the older shape before the reader touched anything.

### Tests

Eleven, in a new `SettingsDurabilityTests`, each mutation-checked:

| mutation | fails |
|---|---|
| drop the converter branch | `A_property_with_its_own_converter_survives_the_salvage` |
| drop the re-seed | `Sanitize_restores_a_missing_script_font_key`, `The_reseed_covers_every_script_…` |
| drop the extension-data member | `Unknown_top_level_properties_survive_a_round_trip` |
| drop the newer-than-supported guard | `A_newer_settings_file_is_not_rewritten_on_load` |
| dispose the logger before the swap / not at all | `Swapping_the_global_logger_…` |

Plus the guards against each fix overshooting: the re-seed is idempotent and never overwrites a font the reader chose; a repairable file this build *does* understand is still written back; an ordinary file gains no stray member in its output.

**#878 has no failing-without-the-fix test** and I would rather say so than imply otherwise: a torn write is a race, so a test cannot reliably reproduce it. `Concurrent_saves_leave_a_file_that_parses` pins that the serialized path is correct under concurrency and stays that way; the absence of the lock is the mechanism, and it was traced rather than reproduced.

Full suite: **2666 passed, 0 failed, 17 skipped**.

### DIFF
--- a/src/CST.Avalonia.Tests/Services/SettingsDurabilityTests.cs
+++ b/src/CST.Avalonia.Tests/Services/SettingsDurabilityTests.cs
@@ -1,0 +1,311 @@
+using System;
+using System.IO;
+using System.Linq;
+using System.Text.Json;
+using System.Threading.Tasks;
+using CST.Avalonia.Models;
+using CST.Avalonia.Services;
+using Xunit;
+
+namespace CST.Avalonia.Tests.Services;
+
+/// <summary>
+/// The settings half of the durability pass: concurrent saves, the salvage path's converter gap, a script
+/// key that went missing, and what an older build does to a newer build's file. (#878, #880, #881, #883)
+/// </summary>
+public sealed class SettingsDurabilityTests : IDisposable
+{
+    private readonly string _dir;
+
+    public SettingsDurabilityTests()
+    {
+        _dir = Path.Combine(Path.GetTempPath(), $"settings-durability-{Guid.NewGuid():N}");
+        Directory.CreateDirectory(_dir);
+    }
+
+    private string SettingsPath => Path.Combine(_dir, "settings.json");
+
+    // ---- concurrent saves (#878) -----------------------------------------------------------------------
+
+    /// <summary>
+    /// Saves fired at once still leave a file that parses.
+    ///
+    /// <para>They shared one <c>settings.json.tmp</c>: A could finish writing the temp file, B truncate and
+    /// start rewriting it, and A's <c>File.Replace</c> promote B's half-written temp over the real file.
+    /// Concurrent callers are ordinary — the 750ms debounce flushes on a pool thread while
+    /// <c>XmlUpdateService</c> and <c>IndexingService</c> save directly from background flows.</para>
+    ///
+    /// <para>A torn write is a race, so this cannot prove the lock by failing reliably without it — what it
+    /// pins is that the serialized path is correct under concurrency and stays that way. The lock's absence
+    /// is the mechanism; the reviewer traced it.</para>
+    /// </summary>
+    [Fact]
+    public async Task Concurrent_saves_leave_a_file_that_parses()
+    {
+        var service = new SettingsService(_dir);
+        service.Settings.XmlBooksDirectory = Path.Combine(_dir, "books");
+
+        await Task.WhenAll(Enumerable.Range(0, 24).Select(_ => service.SaveSettingsAsync()));
+
+        var reloaded = JsonSerializer.Deserialize<Settings>(File.ReadAllText(SettingsPath),
+            new JsonSerializerOptions { PropertyNameCaseInsensitive = true });
+        Assert.NotNull(reloaded);
+        Assert.Equal(service.Settings.XmlBooksDirectory, reloaded!.XmlBooksDirectory);
+        Assert.False(File.Exists(SettingsPath + ".tmp"));
+    }
+
+    // ---- the salvage path's converter gap (#880) -------------------------------------------------------
+
+    /// <summary>
+    /// A property with its own converter is read by that converter on the salvage path too.
+    ///
+    /// <para><c>AiConnectionRecord.Headers</c> declares <c>AiHeaderRecordListConverter</c>, which reads the
+    /// legacy dict-shaped form. Taken apart property-by-property the node is an object where a List is
+    /// expected, so it landed in the drop-the-collection branch and the headers were discarded — the salvage
+    /// path being strictly weaker than the strict path, in the one mechanism whose job is to lose as little
+    /// as possible.</para>
+    ///
+    /// <para>The connection below carries a number where <c>displayName</c> expects a string. That is what
+    /// forces this record to be taken apart at all — <see cref="TolerantSettingsReader"/> tries a whole-node
+    /// read at every level first, precisely so the ordinary converters apply wherever they can, and the
+    /// converter gap only shows once something ELSE on the same object has failed. A fixture broken further
+    /// up the tree does not reach it: the <c>ai</c> node deserializes whole and the headers were never at
+    /// risk.</para>
+    /// </summary>
+    [Fact]
+    public void A_property_with_its_own_converter_survives_the_salvage()
+    {
+        const string json = """
+        {
+          "version": "1.0",
+          "ai": { "chat": { "connections": [ {
+              "id": "mine",
+              "displayName": 12345,
+              "headers": { "X-Org": "acme", "X-Team": "pali" }
+          } ] } }
+        }
+        """;
+
+        var salvaged = TolerantSettingsReader.Read<Settings>(
+            json, new JsonSerializerOptions { PropertyNameCaseInsensitive = true }, out var dropped);
+
+        Assert.NotNull(salvaged);
+        var connection = Assert.Single(salvaged!.Ai.Chat.Connections);
+        Assert.Equal(2, connection.Headers.Count);
+        Assert.Contains(connection.Headers, h => h.Name == "X-Org" && h.Value == "acme");
+        Assert.DoesNotContain(dropped, d => d.Contains("Headers", StringComparison.OrdinalIgnoreCase));
+    }
+
+    // ---- a script key that went missing (#881) ---------------------------------------------------------
+
+    /// <summary>
+    /// Sanitize puts back a canonical script key that is absent.
+    ///
+    /// <para>The Appearance panel builds its rows by enumerating this dictionary, so a key dropped by a
+    /// salvage or a hand-edit left that script with no font control at all — permanently, since nothing ever
+    /// re-seeded it, and invisibly, since rendering falls back safely.</para>
+    /// </summary>
+    [Fact]
+    public void Sanitize_restores_a_missing_script_font_key()
+    {
+        var settings = new Settings();
+        settings.FontSettings.ScriptFonts.Remove("Myanmar");
+        settings.FontSettings.ScriptFonts.Remove("Sinhala");
+
+        var fixes = SettingsValidator.Sanitize(settings);
+
+        Assert.Contains("Myanmar", settings.FontSettings.ScriptFonts.Keys);
+        Assert.Contains("Sinhala", settings.FontSettings.ScriptFonts.Keys);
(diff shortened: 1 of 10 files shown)


fsnow:
## Fable review round

The reviewer verified the runtime claims empirically against the exact Serilog versions the app ships, and **found my #882 fix was wrong**. That was the one I flagged as least sure, and it is the one that mattered.

### HIGH — #882 traded a Windows-only failure for a cross-platform one

Disposing the replaced logger does free the Windows file handle. But an `ILogger` captured at construction is bound to the pipeline it came from — and this app captures **20+ via `Log.ForContext`, plus every MEL `ILogger<T>` the DI container injects** (both confirmed empirically). Disposal silently stops *their* file output on every platform, while the console sink goes on working and hides it. The very next line of the level-changed handler is written through such a logger and would have vanished.

On macOS the old leak-but-keep-alive behaviour at least kept everything logging, just at a stale level. **My fix regressed that.**

The correct answer is the standard Serilog one, and it is simpler than either version: a `LoggingLevelSwitch` created once and wired in with `MinimumLevel.ControlledBy`. Changing the level moves the switch and builds nothing — no second logger, no dispose, no handle contention on any platform. And it does something neither the old behaviour nor my fix managed: **a logger already captured honours the new level immediately**, instead of keeping the one it was built with. `SwapGlobalLogger` is gone.

The replacement test is the one that would have caught this: it logs through a `ForContext` child captured *before* the change and asserts the event still lands.

### MED — the converter attempt caught too narrowly

`SalvageObject`'s new block caught only `JsonException`, while `Node()` catches `NotSupportedException` too for the *identical* `Deserialize` call one branch below ("a converter refusing the shape outright"). `Read()`'s own catch is `JsonException` only, so an NSE would escape and abort the **whole salvage**, falling through to backups — worse than never having tried the converter. No live thrower today, but the block exists for the next converter, which is exactly when it would bite.

### The LOWs, all acted on

- The converter test now asserts its own premise — that `displayName` was dropped. Without it the test passes vacuously the day something makes the record deserialize whole, which is how my *first* version of it passed with the fix reverted.
- A state-side twin of the newer-file-not-rewritten test.
- Three comments corrected to claim only what the code does: the extension-data member's omission comes from System.Text.Json skipping a null extension member, **not** from `WhenWritingDefault` (the settings options set no ignore condition); the flush drain leaves a flag-to-lock window; and the salvage path does not populate extension data, so a newer file that *also* fails a strict read loses its unknown keys there — bounded by the copy-preserve, not fixed.
- Named the pre-existing serialize-under-mutation race beside the new write lock, so it is not mistaken for something the lock covers.

### Confirmed correct as written

#881 in full. #878's lock placement (it covers serialize, temp write, `File.Replace` **and** `WriteBackupAsync` and the `_backupNextSave` read-reset), no re-entrancy, no deadlock on the async-end-to-end shutdown path, no semaphore disposal needed. #880's degradation for converter factories, open generics and missing constructors. #883's `Migrate` interplay and extension-data carry-through on later saves.

### One correction to a claim I made

My PR text said the `#882` test discriminated both halves. After the redesign the first version of the replacement test drove `App.LogLevelSwitch` directly, so it passed with `ReconfigureLogger` gutted. Routed through the real path (`ApplyLogLevel`) and re-checked.

Full suite: **2672 passed, 0 failed, 17 skipped**. Still wants a Windows run to confirm the original symptom is gone.

Issue #910, opened 2026-08-30
Hiding the last tool reclaims its width only if that tool was the Assistant (R9-2, re-derived after #908)


fsnow:
`RebalanceMainDock()` has exactly one caller in the app — `LayoutViewModel.HideAssistantPanel`. The other three panels hide through `HideToolPanel` without it. That was merely asymmetric when the Assistant was its own right-hand column; since #908 moved it into the left tool dock, all four are the same kind of tab in the same dock, and the asymmetry now produces a rule that depends on which tab you closed last.

## Expected

Hiding the last remaining tool empties the left rail, so the documents take the full window width. Which tool happened to be last should not matter — they are four tabs in one dock.

## Actual

The width is handed back only when the last tool was the **Assistant**. `HideAssistantPanel` calls `HideToolPanel(...)` and then `RebalanceMainDock()`; `HideSelectBookPanel`, `ShowDictionaryPanel`'s hide counterpart and the Search hide all call `HideToolPanel(...)` alone.

## Contrast — the same action, differing only in which tool

- Open only the Assistant, hide it → `ApplyMainDockProportions` runs, finds no `LeftTools`/`LeftToolDock`, sets `leftShare = 0` and `documents.Proportion = 1.0`. Correct.
- Open only Search, hide it → nothing restates the row. `MainDocumentDock` keeps the `1.0 - 0.25` it was constructed with, and the freed quarter is left for the framework to resolve.

The rail is removed identically in both cases; only the proportion restatement differs.

## Second half: the recreate path never restates the row either

`CstDockFactory.EnsureLeftToolDock` builds `LeftTools` with `Proportion = 0.25` and inserts it plus `MainSplitter` at the head of `MainDock`, but does not restate `MainDocumentDock`'s share, and nothing on the show path calls `RebalanceMainDock` — `ShowToolPanel` calls `EnsureLeftToolDock` and stops. `ApplyMainDockProportions` is the only code that sets *both* sides.

The two halves compose: hide the Assistant last and documents ends at `1.0`; the next `ShowToolPanel` inserts `LeftTools` at `0.25` into a row already summing to `1.0`.

## Suggested fix

[suggestion] Move the `RebalanceMainDock()` call out of `HideAssistantPanel` and into `HideToolPanel`, so every tool hide is treated alike, and add one on `EnsureLeftToolDock`'s recreate path. Both are the same one-line shape as the existing call.

## How this was found

[...]

PR #911, opened 2026-08-30
Re-derive R9-2 and R9-4 against the single-rail layout

fsnow:
#908 moved the Assistant out of its own right-hand column and into the left tool dock, deleting `EnsureRightToolDock`, the right splitter and rail, and the separate assistant proportion. Both R9 layout findings in the August review were written against the two-rail layout, so neither reads correctly as written.

- **R9-4 (orphan/duplicate `MainSplitter`) is retired.** `MainDock` is now `[LeftTools, MainSplitter, MainDocumentDock]`, so emptying the rail leaves a single non-splitter and `CleanupSplitters`' `<= 1` branch strips the splitter correctly, after which `EnsureLeftToolDock` reinserts exactly one. The review predicted this: *"with the assistant off the ≤1 branch cleans up correctly, which is why it hasn't been seen."*
- **R9-2 survives in a changed and sharper form**, filed as #910. Its original framing — that `EnsureLeftToolDock` lacked the proportion restatement its twin `EnsureRightToolDock` had received — no longer parses, because the twin is gone. What re-derives is that `RebalanceMainDock` has exactly one caller, `HideAssistantPanel`, so hiding the last remaining tool reclaims its width only when that tool happened to be the Assistant — now odd, since all four are tabs in one dock.

Marks both rows in place so neither is picked up and worked from as written, and adds a short note after the R9 table. Doc only; no code change.


fsnow:
Superseded by #912, which carries this same findings-doc commit alongside the #910 fix — per the maintainer's request to commit the doc update together with the code change.

Issue #914, opened 2026-08-31
A passage citation names a paragraph range it does not contain, across a numbering restart (R14-3)


fsnow:
`PassageTool.Describe` renders any multi-paragraph window as `paragraphs {first}-{last}`. Paragraph numbers restart per section in 95 non-Multi books, and per sub-book in Multi books, so a window that crosses a restart produces a continuous-looking range naming text it does not contain.

## Expected

`NormalizedReference` is the app's own attestation of a window's scope — surface B renders it beside a generated answer (#602). If it names paragraphs 55 through 57, the window contains paragraphs 55, 56 and 57.

## Actual

`TeiPassageReader.ReadWindow` is not div-bounded, so a window can begin at ¶55 near the end of one section and end at ¶57 of the next, having passed through that section's ¶1, ¶2, ¶3… The citation reads `paragraphs 55-57`, describing three paragraphs where the window holds dozens, most of them numbered outside the stated range.

`Describe` cannot detect this: it receives only `ParagraphNumber`, `ParagraphBookCode` and `EndParagraphNumber`. It never sees `EndParagraphBookCode`, so a sub-book change does not reach it either, and for a Multi book it labels the whole span with the *start* code alone.

An API consumer has no better recourse: `ParagraphNumber: 55` and `EndParagraphNumber: 57` are equally silent about the restart.

## Contrast — three ways

- **The reversed case already looks wrong.** A restart landing on a lower number gives `paragraphs 289-3`, which a reader distrusts on sight. The overlapping case gives `paragraphs 55-57`, which reads as ordinary and is just as false. The finding is only about the plausible one.
- **The page citation solves the same problem honestly.** A window crossing a page break gets `PagesAcross`, which lists *every* page it covers, in reading order, with the one case where `[0]` is not the start page documented at the source. Paragraph numbering collapses to two endpoints and a hyphen.
- **The window already knows.** `Materialize` holds `readStart` and `end` and has `markers` in hand, so the paragraph entries actually spanned are available at the point the citation is built. Nothing consults them.

## Suggested fix

[...]

Issue #310, opened 2026-07-12
[HIGH] Snippet/passage corruption: a range starting inside a <note> leaks apparatus + drops the rest of the window


fsnow:
`TeiText.Clean`'s note-strip branch (`TeiText.cs:44`) matches `name == "note"` for **both** `<note>` and `</note>` — `TagName("</note>")` returns `"note"` (it skips the leading `/`, `:129`). So when a rendered range **begins inside a note's content**, the first tag hit is the close `</note>`, `SkipSubtree` is entered with depth 1 and consumes to the *next* `</note>` — or the entire rest of the window. Verified empirically: cleaning from inside `<note>variant one</note> ccc ddd <note>variant two</note> eee` returns `"ant one"` — note tail rendered as base text, all following base text silently dropped.

**Reachable paths (note text IS indexed — `DevaXmlTokenizer.FirstPass` doesn't strip notes):**
- **Hit inside a note** → `Extract`'s suffix segment `Clean(mark.End, winEnd)` starts mid-note → snippet shows the note tail and loses the post-note suffix.
- **Long-sentence base hit** → `ProseWindow`'s `SnapToSpace` can snap to a space *inside* a note; the tag-nudge only fixes bounds inside a **tag**, not inside a note **element**.
- **Passage reader** → `SnapBackToSentenceStart`/`WalkBackward` are NOT note-aware (unlike the snippet extractor's `SentenceStart/End`), so a snapped start/prev-cursor can land mid-note and `WalkForward` then skips to the next `</note>`, silently jumping the passage.

**Fix:** (a) in `Clean`/`WalkForward`, treat a close tag (`tag.StartsWith("</")`) in the strip branch as a zero-width tag, never a subtree; (b) nudge window bounds out of note regions (via `NoteRegions`/`InNote`) like they're nudged out of tags; (c) make `TeiPassageReader`'s sentence scans note-aware. Related: `NoteCount` (A4-15) should count notes *intersecting* the window, not just those starting in it. (A4-2, A4-15)

From the Fable-led code review (area A4); full triaged detail in docs/development/AI_INTEGRATION_REVIEW_FINDINGS.md. Part of epic #186.

PR #916, opened 2026-08-31
Keep a passage window honest about its text and its scope

### LINKED ISSUE #913: A passage window can open mid-note and render the footnote's tail as base text (R14-1)
`TeiPassageReader.ReadWindow` never moves its start out of a `<note>`. When the window opens inside one, `TeiText.Clean` sees only the closing `</note>` — zero-width since #310 — and emits the note's tail (a variant reading plus its sigla) as undelimited base text.

## Expected

A passage window renders base text as base text and apparatus as apparatus, wherever its bounds happen to fall. Note text is indexed, so a cursor pointing into a note is a normal thing to receive.

## Actual

`ReadWindow` sets `readStart = startPos` and only ever moves it via the optional sentence snap. That snap is rejected when `WalkForward(candidate, maxChars, …) <= startPos` — i.e. whenever the base text between the enclosing sentence start and the note exceeds roughly `1.5 x maxChars` rendered characters — and it does not run at all unless `snapStartToSentence` is set. Either way `readStart` stays inside the note, and the window opens mid-apparatus.

With `structuredNotes: true` it is worse in a quieter way: `TeiText.SplitBracedNotes` sees a closing `}` with no opening `{` and passes it through, so a stray brace lands in the text documented as clean and quotable.

## Contrast — the sibling surface gets this right

`TeiSnippetExtractor` handles the identical hazard: it computes `TeiText.NoteRegions` for its window and pushes each bound out of any note (`NudgePastNote` for the start, `NudgeBeforeNote` for the end) before rendering, clamped so a hit inside a note still shows. Same corpus position, same note, two surfaces, different answers — and the passage reader is the one an agent quotes from.

The reader is not unaware of notes: its own sentence snap is note-aware, using `NoteRegions` so a danda inside a note is not mistaken for a sentence boundary. It just never applies that knowledge to the start it actually uses.

## Trigger

`maxChars` is caller-controlled and `PassageTool` clamps it to `1..20000`, so a small budget makes this reliable rather than rare: `passage(cursor: <hit-in-note>, maxChars: <=100)` leaves the start inside the note every time. At the default 1200 budget it needs a sentence longer than ~1800 characters before the note, which the corpus does contain.

## Suggested fix

[suggestion] Mirror the snippet extractor: when `readStart` falls inside a note region, move it to the note **start** rather than its end, so the content the cursor pointed at is still rendered and is rendered as a note. Guard it the way the sentence snap is guarded — if starting at the note start cannot produce a window reaching past `startPos`, move to the note end instead, which always advances and so cannot make a paging cursor loop.

## How this was found

[observed] Area R14 of the August 2026 Fable code review, finding **R14-1**, recorded in `docs/development/CODE_REVIEW_2026-08_FINDINGS.md`. Re-derived against `main` at `8067c79`; the mechanism is unchanged by #871, which altered where windows *end*, not where they start.

[observed] It is the last open residual of A4-2/#355. The review's July-fix audit records A4-2 as otherwise intact — "close-tag-zero-width in `Clean`, note-aware sentence scans, snippet nudges — R14-1 above is the one remaining gap on the *passage* start."

[fsnow] Asked for the R14 passage-fidelity pair to be filed and fixed.

Unlabelled.

fsnow:
Fixes #913 and #914 — the passage-fidelity pair from area R14 of the August review. Both are ways a window could describe itself falsely, in the surface an agent quotes from.

## #913 — the window could open mid-note and render the apparatus as base text

`ReadWindow` never moved its start out of a `<note>`. Opening inside one, `TeiText.Clean` meets only the closing tag — zero-width since #310 — so the variant reading and its sigla arrive as undelimited base text. With `structuredNotes` the same start leaves an unmatched `}` in the text documented as clean and quotable, because `SplitBracedNotes` never saw the opening brace.

**Reproduced before fixing.** With a cursor mid-note and `maxChars: 20`, the window's text came back as `"sigla charlie delta।"` — the apparatus leading the passage.

Note text is indexed, so a cursor pointing into a note is ordinary, and `maxChars` is caller-controlled down to 1, which makes a small budget reach this reliably rather than rarely. The existing sentence snap could not stand in for the fix: it is optional, and it declines whenever the base text before the note outruns the budget.

The start now moves to the note's **start**, so what the cursor pointed at is still in the window and is rendered *as* apparatus. It falls back to the note's end only where opening at the start could not produce a window reaching past the cursor — which would let a paging cursor re-read what it just returned and never advance, the same loop the sentence snap already guards against.

`TeiSnippetExtractor` has nudged its bounds out of note regions all along. This is the passage reader adopting its sibling's fix — the review's own dominant pattern.

## #914 — the citation could name a paragraph range the window does not contain

Any multi-paragraph window was cited as `paragraphs {first}-{last}`. Numbers restart per section in 95 non-Multi books and per sub-book in Multi books, so a window can open at ¶55 near the end of one section and close at ¶57 in the next, having passed through that section's own ¶1, ¶2, ¶3. The citation named three paragraphs where the window held four, two of them numbered outside the stated range — and read as entirely ordinary while doing it. The reversed case (`paragraphs 289-3`) at least looks wrong; this one is the dangerous one.

- `BookMarkers.ParagraphsRunContiguously` answers **positionally**, walking the entries rather than comparing the endpoints, because 55 < 57 is exactly the case that must still fail. It compares each entry against the previous one's `Last`, so a ranged paragraph ("16-26") followed by 27 still reads as ascending.
- `PassageWindow.ParagraphsContiguous` and `PassageResult.ParagraphsContiguous` carry the answer, so an API consumer reading the structured fields can detect the restart too — the finding noted `ParagraphNumber: 55` / `EndParagraphNumber: 57` were equally silent.
- `Describe` emits a range only where the run holds, and otherwise names both ends with **their own** book codes and says outright that it is not a continuous range. That also retires the smaller lie the finding mentions: labelling a whole Multi-book span with the start sub-book code alone.

## Tests

Nine new tests in `PassageFidelityTests`, each half mutation-tested:

| Mutation | Fails |
|---|---|
| Remove the note nudge | the two note tests, and only those |
| `Describe` ignores contiguity | the citation test, and only that |
| Contiguity check ignores the number sequence | the two restart tests, and only those |

Full suite: **2716 passed, 0 failed, 17 skipped.**

Test XML follows the house style for this engine — ASCII placeholder words, `।` danda escapes, `Script.Devanagari` output so conversion is the identity and text can be asserted exactly.

## Also

Marks both rows in `CODE_REVIEW_2026-08_FINDINGS.md` as fixed and adds them to the status block. Both were fixed as the review prescribed.


fsnow:
**Claude Code review** — 2 finding(s)

### normal — src/CST.Avalonia/Services/Tools/PassageTool.cs:187-188
The single-paragraph guard `if (endNumber is not int last || last == number) return Name(number.Value, bookCode);` at PassageTool.cs:187-188 checks only the number, not the book code — so in a Multi book, a window that opens at ¶N of one sub-book and closes at ¶N of the next (paragraph numbers restart per sub-book) is cited as "paragraph N (an5)", silently dropping the end sub-book code and the `contiguous=false` signal. This is precisely the class of failure the PR was written to retire ("labelling a whole Multi-book span with the start sub-book code alone"), but the equal-numbers subcase still triggers it. Fix: check both number AND book code before collapsing (`last == number && string.Equals(endBookCode, bookCode, StringComparison.Ordinal)`), or check `!contiguous` first.

### nit — src/CST.Core/Search/TeiPassageReader.cs:46-68
The new note nudge at TeiPassageReader.cs:55-68 chooses `noteStart` when the simulated `WalkForward` advances past the cursor, but with `includeVariants=true, structuredNotes=false` that walk enters the note body and `WalkForward`'s boundary check at line 733 is NOT note-aware (unlike its `WalkBackward`/`ExtendToSentenceEnd`/`SnapBackToSentenceStart` siblings, which all guard with `!TeiText.InNote`). If a danda falls inside the note past the budget position, or if the note body exceeds `hardCap` with no boundary, the walk returns a mid-note position; `TeiText.Clean` then emits the opening `<note>` as `{` but never sees `</note>`, so the returned Text carries an unmatched `{`. This is a narrow residual (specific mode + specific note shape) of a strictly-improving fix, not a regression — the pre-PR path returned a different malformation (unmatched `}`) at the same call — so it does not need to block; a follow-up could either make the `WalkForward` boundary check note-aware for consistency with its siblings, or reject `noteStart` when the simulated walk lands mid-note (falling back to `noteEnd`).




---
_Generated by [Claude Code](https://claude.ai/code)_

fsnow:
Both ultra-review findings are fixed in `01e46a3`.

**Finding 1 (normal) — the equal-numbers collapse.** Correct, and it was the exact failure this PR claimed to retire, surviving in a subcase I did not think to test. `Describe` collapsed to one end whenever `last == number`, before consulting `contiguous` at all — so a Multi-book window opening at ¶N of one sub-book and closing at ¶N of the next was cited as `paragraph N (an5)`, dropping both the far sub-book's code and the break. The contiguity test now runs first; equal numbers collapse only on a straight run.

**Finding 2 (nit) — the mid-note end.** Verified before acting on it: `WalkForward` is indeed the only `IsBoundary` call in `TeiPassageReader` without an `!InNote` guard, where `WalkBackSentences`, `ExtendToSentenceEnd`, `WalkForwardSentences` and `SnapBackToSentenceStart` all have one. The nudge now requires the whole note to *fit* before opening a window at a note's start, and skips the note otherwise.

**And it was not theoretical.** The new test fails with the fit condition removed, so the unmatched opening brace was genuinely reachable — the mirror image of the closing brace this PR was written to stop.

I took the narrower of the two options offered and filed the broader one as **#917**: making `WalkForward`'s check note-aware restores consistency with its four siblings, but it moves where *every* passage window ends rather than only those touching notes, and it shares a return path with #871's akṣara retreat — so it deserves its own change and its own review rather than riding along here. What is fixed in this PR is the path this PR itself opened.

Both mutation-tested: restoring the equal-numbers collapse fails the sub-book test alone; dropping the whole-note-must-fit condition fails the long-note test alone. Full suite 2718 passed, 0 failed, 17 skipped.

PR #981, opened 2026-09-03
The five books with no page markers are expected, not a gap

fsnow:
Follow-up to #980. **[fsnow]**, on whether the absence needed confirming: *"not our concern"*.

The table I just merged closed with an open question: *"Worth confirming this is a property of the printed sources rather than a gap in the XML."* That was not an open question, and I should not have written it as one.

The five books are `e*` texts. The VRI printed set does not extend to them, so there is no printed pagination for them to carry — in any edition. It is the same reason Añña coverage is thin generally: 5% VRI, 70% Myanmar, no PTS or Thai at all.

Leaving the question in the doc invites the next reader to go hunting for something that was never missing. That is the more expensive mistake here: the corpus is VRI's, and an agent chasing a phantom data gap in it is the kind of investigation that reaches the wrong people.

One paragraph in the generator, regenerated.


fsnow:
Closing unmerged. **[fsnow]:** *"Your initial statement was more true than the correction. The point is that this is an issue for the Tipitaka.org team, not for CST Reader."*

The doc's original line stands: whether those five books reflect the printed sources or a gap in the XML is a real question. What I got wrong was the routing, not the phrasing — it is a corpus question for the Tipiṭaka.org team, and not one CST Reader answers or explains away.

